PR #30, opened 2015-02-19
support for markdown output

Friz-zy:
Goals: add markdown as output format because readthedocs has support only for rst and md formats right now. 
- add cli options: --markdown, --walk-path, --walk-path-followlinks
- change cli option --html-dir to --output-dir
- add md.mako template based on text.mako
- you can now specify many modules per one executing
- walk-path option allows recursively find all modules in path via **init**.py file, '.py' extension and even via 'python' word in the shebang

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -353,11 +353,9 @@ def import_module(module_name):
         # contained in `pdoc.import_path`.
         imp.find_module(module_name.split('.')[0], import_path)
 
-    if module_name in sys.modules:
-        return sys.modules[module_name]
-    else:
+    if module_name not in sys.modules:
         __import__(module_name)
-        return sys.modules[module_name]
+    return sys.modules[module_name]
 
 
 def _source(obj):
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -208,6 +208,18 @@ html_package_name = 'index.html'
 The file name to use for a package's `__init__.py` module.
 """
 
+markdown_module_suffix = '.m.md'
+"""
+The suffix to use for module markdown files. By default, this is set to
+`.m.md`, where the extra `.m` is used to differentiate a package's
+`index.md` from a submodule called `index`.
+"""
+
+markdown_package_name = 'index.md'
+"""
+The file name to use for a package's `__init__.py` module.
+"""
+
 import_path = sys.path[:]
 """
 A list of paths to restrict imports to. Any module that cannot be
@@ -289,6 +301,27 @@ def text(module_name, docfilter=None, allsubmodules=False):
                  docfilter=docfilter,
                  allsubmodules=allsubmodules)
     return mod.text()
+
+
+def markdown(module_name, docfilter=None, allsubmodules=False):
+    """
+    Returns the documentation for the module `module_name` in
+    markdown format. The module must be importable.
+
+    `docfilter` is an optional predicate that controls which
+    documentation objects are shown in the output. It is a single
+    argument function that takes a documentation object and returns
+    True of False. If False, that object will not be included in the
+    output.
+
+    If `allsubmodules` is `True`, then every submodule of this module
+    that can be found will be included in the documentation, regardless
+    of whether `__all__` contains it.
+    """
+    mod = Module(import_module(module_name),
+                 docfilter=docfilter,
+                 allsubmodules=allsubmodules)
+    return mod.markdown()
 
 
 def import_module(module_name):
@@ -646,6 +679,14 @@ class Module (Doc):
         Returns the documentation for this module as plain text.
         """
         t = _get_tpl('/text.mako')
+        text, _ = re.subn('\n\n\n+', '\n\n', t.render(module=self).strip())
+        return text
+
+    def markdown(self):
+        """
+        Returns the documentation for this module as plain markdown.
+        """
+        t = _get_tpl('/md.mako')
         text, _ = re.subn('\n\n\n+', '\n\n', t.render(module=self).strip())
         return text
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -208,7 +208,7 @@ html_package_name = 'index.html'
 The file name to use for a package's `__init__.py` module.
 """
 
-markdown_module_suffix = '.m.md'
+markdown_module_suffix = '.md'
 """
 The suffix to use for module markdown files. By default, this is set to
 `.m.md`, where the extra `.m` is used to differentiate a package's
--- a/pdoc/templates/md.mako
+++ b/pdoc/templates/md.mako
@@ -2,6 +2,8 @@
 
 <%!
   import re
+  from os.path import join
+  from pdoc import markdown_module_suffix, markdown_package_name
 
   def docstring(d):
       if len(d.docstring) == 0 and hasattr(d, 'inherits'):
@@ -71,6 +73,20 @@
           strings.append('```')
           codeblock = False
       return '\n'.join(strings)
+
+  def submodule_link(submodule, module):
+      path = submodule.name
+      path = path.replace(module.name, '', 1)
+      path = path.lstrip('.')
+      path = path.split('.')
+      if submodule.is_package():
+          path.append(markdown_package_name)
+      else:
+          e = path[-1]
+          del path[-1]
+          path.append(e + markdown_module_suffix)
+      path = join(*path)
+      return path
 %>
 
 <%def name="function(func)" filter="trim">
@@ -201,7 +217,7 @@ ${class_(c)}
 Sub-modules
 -----------
 % for m in submodules:
-- ${m.name}
+- [${m.name}](${submodule_link(m, module)})
 
 % endfor
 % endif
--- a/pdoc/templates/md.mako
+++ b/pdoc/templates/md.mako
@@ -1,0 +1,178 @@
+## Define mini-templates for each portion of the doco.
+
+<%!
+  import re
+
+  def docstring(d):
+      if len(d.docstring) == 0 and hasattr(d, 'inherits'):
+          return d.inherits.docstring
+      else:
+          return d.docstring
+
+  def nl2br(str):
+      str = str.replace("\r\n\r\n", "\r\n\r\n    ")
+      str.replace("\r\n", "  \r\n")
+      str = str.replace("\n\n", "\n\n    ")
+      str.replace("\n", "  \n")
+      str = str.replace("\r\r", "\r\r    ")
+      str.replace("\r", "  \r")
+      return str
+
+  def h1(str):
+      str = "# " + str
+      return str
+
+  def h2(str):
+      str = "## " + str
+      return str
+
+  def h3(str):
+      str = "### " + str
+      return str
+
+  def h4(str):
+      str = "#### " + str
+      return str
+
+  def h6(str):
+      str = "##### " + str
+      return str
+
+  def h6(str):
+      str = "###### " + str
+      return str
+
+%>
+
+<%def name="function(func)" filter="trim">
+**${func.name}** (${func.spec()})
+
+    ${docstring(func) | nl2br}  
+
+</%def>
+
+<%def name="variable(var)" filter="trim">
+**${var.name}**
+
+    ${docstring(var) | nl2br}  
+
+</%def>
+
+<%def name="class_(cls)" filter="trim">
+${cls.name | h4} \
+% if len(cls.docstring) > 0:
+
+${cls.docstring}  
+
+% endif
+<%
+  class_vars = cls.class_variables()
+  static_methods = cls.functions()
+  inst_vars = cls.instance_variables()
+  methods = cls.methods()
+  mro = cls.module.mro(cls)
+  descendents = cls.module.descendents(cls)
+%>
+% if len(mro) > 0:
+${"Ancestors (in MRO)" | h6}
+% for c in mro:
+- ${c.refname}
+
+% endfor
+% endif
+
+% if len(descendents) > 0:
+${"Descendents" | h6}
+% for c in descendents:
+- ${c.refname}
+
+% endfor
+% endif
+
+% if len(class_vars) > 0:
+${"Class variables" | h6}
+% for v in class_vars:
+- ${capture(variable, v)}
+
+% endfor
+% endif
+
+% if len(static_methods) > 0:
+${"Static methods" | h6}
+% for f in static_methods:
+- ${capture(function, f)}
+
+% endfor
+% endif
+
+% if len(inst_vars) > 0:
+${"Instance variables" | h6}
+% for v in inst_vars:
+- ${capture(variable, v)}
+
+% endfor
+% endif
+
+% if len(methods) > 0:
+${"Methods" | h6}
+% for m in methods:
+- ${capture(function, m)}
+
+% endfor
+% endif
+</%def>
+
+## Start the output logic for an entire module.
+
+<%
+  variables = module.variables()
+  classes = module.classes()
+  functions = module.functions()
+  submodules = module.submodules()
+%>
+
+Module ${module.name}
+-------${'-' * len(module.name)}
+% if not module._filtering:
+${module.docstring}
+% endif
+
+
+% if len(variables) > 0:
+Variables
+---------
+% for v in variables:
+- ${variable(v)}
+
+% endfor
+% endif
+
+
+% if len(functions) > 0:
+Functions
+---------
+% for f in functions:
+- ${function(f)}
+
+% endfor
+% endif
+
+
+% if len(classes) > 0:
+Classes
+-------
+% for c in classes:
+${class_(c)}
+
+% endfor
+% endif
+
+
+% if len(submodules) > 0:
+Sub-modules
+-----------
+% for m in submodules:
+- ${m.name}
+
+% endfor
+% endif
--- a/pdoc/templates/md.mako
+++ b/pdoc/templates/md.mako
@@ -13,11 +13,15 @@
 
   def nl2br(string):
       string = string.replace("\r\n\r\n", "\r\n\r\n    ")
-      string.replace("\r\n", "  \r\n")
+      string = string.replace("\r\n  \r\n", "\r\n  \r\n    ")
+      #string.replace("\r\n", "  \r\n")
       string = string.replace("\n\n", "\n\n    ")
-      string.replace("\n", "  \n")
+      string = string.replace("\n  \n", "\n  \n    ")
+      #string.replace("\n", "  \n")
       string = string.replace("\r\r", "\r\r    ")
-      string.replace("\r", "  \r")
+      string = string.replace("\r  \r", "\r  \r    ")
+      #string.replace("\r", "  \r")
+      string = string.replace("    ```", "```")
       return string
 
   def h1(string):
@@ -36,7 +40,7 @@
       string = "#### " + string
       return string
 
-  def h6(string):
+  def h5(string):
       string = "##### " + string
       return string
 
@@ -46,18 +50,21 @@
 
   def makeup(string):
       strings = string.splitlines()
+      p = re.compile(r':\s*$')
       dd = re.compile(r':')
       li = re.compile(r'^([0-9]+[\.)]|\*|\-\+)')
       snc = re.compile(r'^[A-Z]\w?')
       code = re.compile(r'^(#!|\.{1,3}|>{1,3})')
-      qu = re.compile(r'^`{1,3}')
+      qu = re.compile(r'^`{3}')
       codeblock = False
       for i, s in enumerate(strings):
           s = s.lstrip()
           if snc.search(s) and i != 0:
               strings[i-1] += '  '
-          if dd.search(s):
+          if p.search(s):
               s += '  '
+          elif  dd.search(s):
+              s = '\n* ' + s
           if li.search(s):
               s = '\n' + s
           if qu.search(s):
@@ -119,7 +126,7 @@ ${cls.docstring | makeup}
   descendents = cls.module.descendents(cls)
 %>
 % if len(mro) > 0:
-${"Ancestors (in MRO)" | h6}
+${"Ancestors (in MRO)" | h5}
 % for c in mro:
 - ${c.refname}
 
@@ -127,7 +134,7 @@ ${"Ancestors (in MRO)" | h6}
 % endif
 
 % if len(descendents) > 0:
-${"Descendents" | h6}
+${"Descendents" | h5}
 % for c in descendents:
 - ${c.refname}
 
@@ -135,7 +142,7 @@ ${"Descendents" | h6}
 % endif
 
 % if len(class_vars) > 0:
-${"Class variables" | h6}
+${"Class variables" | h5}
 % for v in class_vars:
 - ${capture(variable, v)}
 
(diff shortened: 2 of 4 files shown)


BurntSushi:
This looks awesome, thank you.

One small nit: could you please deprecate the `--html-dir` option? By that I mean, keep the flag enabled, and if it's set, output an error message saying to use `--output-dir` instead.

Friz-zy:
Done
All new fixes and features will be pushed in new pull requests

BurntSushi:
Thank you for the ping! I just want to try this out first before merging because it's a big patch. Should hopefully get it merged this weekend.

Friz-zy:
Of course, let me know if something is wrong

Friz-zy:
any updates?

gramster:
I've used this; it's a great addition. It would be nice to have some more control over styling (perhaps it exists and I haven't seen it). There seem to be some issues with the things I have tried although perhaps these are pdoc issues rather than issues with this PR.

For example, I am finding that the Return section is indented from the Args section rather than at the same level, and that the next method's documentation starts immediately after the previous one with no whitespace in between. It would be good to have some whitespace and to have the method names/signatures emphasized, through a larger font/bold or some other way.

That said this is a very valuable contribution; thanks!

Friz-zy:
Thanks for your feedback!

I'll planning proceed with it (and also pull all new pull requests from another contributors to my repo) and share it in my local area on python meetup.

Main problem that now i plays with markdown mako template and if you want you can add some logic and send a request to my [repo](https://github.com/Friz-zy/pdoc).

jvantuyl:
FYI, this is also very useful for MakeDocs-generated documentation.  +1

BurntSushi:
As I mentioned in #99, I'm concerned about aspects of this PR. I don't think `pdoc` should be trying to read random configuration files, and I'm skeptical of the huge diff in general to `scripts/pdoc`.

I'm sorry I've taken so long to give good feedback on this, but I think if it were scaled back considerably then it might help me understand it faster.

jvantuyl:
:(

I agree to some extent about the config file concern. That said, in the absence of a plugin system, I don't see how not to bake this in.

In the long run, pdoc has two pieces of functionality:

* documentation extractor
* static site generator

There is a profound need in the Python community for the former and about a dozen entrenched (and valuable) implementations of the latter.

This pull exists to bridge pdoc into mkdocs which bridges fairly cleanly into readthedocs. My primary use case is that the Python community has sorely neglected automated API docs and I really need this so I don't have to maintain two sets of docs (in Sphinx and my code).

Would you entertain something that doesn't do full mkdocs integration? Would you prefer some other plugin architecture? How can we make this happen, because epydoc doesn't cut it anymore and embedding pdoc-generated HTML is really incongruous.

BurntSushi:
I am in principle OK with outputting Markdown, but it should be done in exactly the same way that HTML or plain text is output today.

I'm not familiar with `mkdocs` and have had pretty bad experience with RTD (which is not to say that I think RTD is bad, but I think Python, and in a broader sense, its packaging system, makes what RTD is trying to do really hard). Could you perhaps elaborate in more detail on the problem you're trying to solve? (Keeping in mind that I've never used `mkdocs`.)

jvantuyl:
Story Time:

You're an engineer. You've used Python extensively. You convinced a team of developers to turn your rag-tag bundle of scripts into a bonafide product.

It's going to be used by developed outside your team.  Maybe it's open source, even. That creates many considerations.

One of this is that you need to produce docs.  You need to generate API docs in addition to other freeform docs. You think, "This must be a solved problem."

You ask the various novices. They point you at ReadTheDocs and GitHub Pages, because that's what all the open source projects use.

GitHub Pages is a bust. It just hosts html or uses a neutered Jekyll implementation. Finding a web host wasn't an issue. The Jekyll support is, we'll Jekyll. Liquid templates are weird and navigation is primitive. Many of the best plugins are unsupported. The complete lack of API docs is killer. You note a few people rendering them to HTML and dumping that into a directory and find it most unsatisfying and the integration is poor.

You head towards ReadTheDocs note that it supports Sphinx and ReadTheDocs.  You dive into Sphinx and immediately notice a heavily-overloaded ReST syntax. You also notice how braindead it is about generating links into the various pages. You briefly become excited but automodule and autofunction only to dig into api-autogen and recoil in horror.

You approach the Python wizards with the hope that you're just missing something. They squint at you, wave you back in the direction of Sphinx, and go back to arguing over who gets to use the chopsticks to eat their noodles.

You discover pdoc. It's perfect. It eschews templates, but the HTML is almost perfect to embed yourself. After fighting with transforming it or patching it for a while, you wonder at why it just doesn't drop HTML fragments or even just markdown.

Markdown! Oh! You remember mkdocs. It seemed really limited but you recall that RTD supports it and it's how somebody recommended doing markdown.

It works great! Now we just need to get the pdoc stuff in there.

So that's where I find myself. Also, I wouldn't be opposed to outputting Sphinx-friendly ReST. Maybe this needs a plugin system?

BurntSushi:
I'm not adding a plugin system. The whole point of `pdoc` is to keep it simple. Plugin systems are not simple. I wrote it explicitly because Sphinx is *not* simple. `pdoc` is my response to exactly the reaction you have to Sphinx (because I had it too). I persisted with `epydoc` for years, and I probably would have continued, but it doesn't support Python 3.

With that said, here are a few points:

1. `pdoc` does use templates. It uses mako. They're here: https://github.com/BurntSushi/pdoc/tree/master/pdoc/templates You can even customize them and specify a new directory to read templates from with `pdoc --template-dir`.
2. Please, help me help you. Could you describe the problem you're trying to solve? **I've never used `mkdocs`, so I don't know what it is you're trying to do.** I'm looking for details, here. What is it about pdoc's HTML that is unsuitable? What do you need? What is mkdocs doing for you? How would dumping to Markdown solve the problem? Why does this need to read a `mkdocs` config file?
3. I suggest avoiding debating the goals of this project. I'm the one who has to maintain it, and my only hope of maintaining it is to keep it simple. Instead, let's try to focus on the actual problem you're having.

jvantuyl:
Most concretely, I want to put the output of `pdoc` into `mkdocs` and I want all of the features to "just work."  I want `pdoc` to be able to discover where to generate things based on the extant `mkdocs` configuration... I shouldn't have to repeat myself.  I want `pdoc` to generate the appropriate metadata so it fits into a hierarchical `mkdocs` navigation.  I want `pdoc` to generate sufficient metadata so I can index it with `mkdocs` RTD theme's `lunr.js` search implementation.  I want to be able to generate something that will be styled like the surrounding documentation.

In short, I want API-docs not to be a wart on every docs site because the people that write generators for them refuse to make them either extensible or output documentation in a way that's composable.  That said, I don't really want `pdoc` to necessarily do any of this.  I want this stuff to all be possible without maintaining a gigabyte of fragile wrapper scripts and transformation code.

Critically, I'm not debating the goals of your project.  If anything, what I want to do is keep `pdoc` simple and focused.  I do not believe that everything I want is appropriate to shoehorn into `pdoc`.  I want `pdoc` to generate documentation, simply, with minimal dependencies.  However, I want the *output* of `pdoc` to be a first-class citizen in a larger world.

More generally, I want to generate docs and include them in others things.  The current text and HTML implementations falls well short of that.  Markdown is a great example of a door-opener.  Sphinx is widely used, so ReST would be lovely, too.  There will be people who want LaTeX I'm sure.  I could go on, but there are about a hundred things you could support and I suspect that you won't want to because you (quite reasonably) don't want to add that complexity.

The way I see it, the minimal work to keep `pdoc` lightweight but still crack it open far enough to get the (wonderfully parsed) docs out would be:

**Option 1:** A plugin architectures does that.  In fact, your entire reason that you rejected this pull request is because, without a plug-in architecture, the only way to get these sorts of features out of it is to introduce unreasonable complexity into `pdoc`.  Plug-in systems not being simple is, well, a significant oversimplification.  Just to see how hard it would be, I wrote one in three hours.  Two hours of that was refactoring the existing code to untangle the `main` behavior a bit.  It was 41 lines (including 17 lines of demo plugin) and introduces zero dependencies.  You can check it out at [here](https://github.com/jvantuyl/pdoc/commit/0f2d1035722d6a18646fc787523b1c974cb4d68b).

**Options 2:** Output some neutral format.  Markdown might be it.  Bare HTML without a layout might be another.  JSON serialization might be good, too.  This would meet my needs, but I think it's a bit overkill and requires a bunch of serialization where access to the module definition would be easier.

BurntSushi:
@jvantuyl I'm asking for specifics because I don't understand the problem you're trying to solve. **I still don't understand the problem you're trying to solve.** Saying, "I want everything to work" or "add a plugin architecture" or "lunr.js search" are not problem statements that I understand. Let's try to avoid proscribing solutions before we understand the problem. I will try to be more specific:

> Most concretely, I want to put the output of pdoc into mkdocs and I want all of the features to "just work." 

What features specifically? Let's say `pdoc` added a feature to output Markdown in the same way that it outputs HTML and plain text. If you had that, what would the sequence of steps be to get what you want? Shell commands might be helpful here.

> I want this stuff to all be possible without maintaining a gigabyte of fragile wrapper scripts and transformation code.

Please, let's refrain from hyperbole.

> More generally, I want to generate docs and include them in others things. The current text and HTML implementations falls well short of that. Markdown is a great example of a door-opener. Sphinx is widely used, so ReST would be lovely, too. There will be people who want LaTeX I'm sure. I could go on, but there are about a hundred things you could support and I suspect that you won't want to because you (quite reasonably) don't want to add that complexity.

You're right, I don't want to support ReST or LaTeX. There are many other tools that are capable of translating, say, Markdown to ReST or LaTeX.

> A plugin architectures does that. In fact, your entire reason that you rejected this pull request is because, without a plug-in architecture, the only way to get these sorts of features out of it is to introduce unreasonable complexity into pdoc.

This PR contains *significantly* more than just "Markdown output." It changes CLI options, adds new directory iteration features, allows one to specify more modules and changes the semantics of how modules are discovered. On top of all that, it tries to read some random config file. All of this is orthogonal to just simply outputting Markdown.

Please help me understand the problem before suggesting a solution.

> Options 2: Output some neutral format. Markdown might be it. Bare HTML without a layout might be another. JSON serialization might be good, too. This would meet my needs, but I think it's a bit overkill and requires a bunch of serialization where access to the module definition would be easier.

JSON serialization is an interesting idea, but probably a change too big than I currently have time for.

mikecharles:
@jvantuyl GitHub's not letting me add a :+1: to your Story Time post, but I agree wholeheartedly. I've been trying to find a solution to the same issues. `pdoc` makes it beautifully simple to get API documentation out of my code, but mkdocs and RTD let me host all of my documentation together, is well known and supported, etc.

I guess just giving a specific example of real markdown files structured for mkdocs or RTD would be helpful. I don't have that because I haven't fully used them yet. That might help @BurntSushi?

BurntSushi:
@mikecharles Specific examples would help a lot.

jvantuyl:
**Specific Example:**  I would like `pdoc` output to happen in Restructured Text, three flavors of Markdown (Standard Markdown, GitHub-flavored Markdown, and with a tables extension), LaTeX, and Encapsulated PostScript so as to be trivial to embed in `mkdocs`, (by extension) ReadTheDocs, (or alternatively) Sphinx, (to generate) LaTeX, and (also) PDFs.

**Implementation Discussion**:

The above is quite literally the documentation target for the project I'm working on right now.

I want to be able to get the docs you generate out without parsing HTML or text.  It's the most fragile way to do it.  Moreover, I don't want you to have to become an expert in six different text formatting languages or to embed ridiculous levels of complexity into `pdoc`.

The two ways I can see to get the docs out of `pdoc` and into something other than a lonely, unloved HTML page that's loosely tied into other documentation are for you to either expose the internals or to produce a profoundly neutral serialization format for them.

We could try to come up with some JSON structure that captures everything.  It would require a lot more talking, a more complex implementation, and some degree of maintenance on your part.  It would be more work on my part.

What would be the least painful is, again, [a plugin architecture using Python's setuptools entry_points](http://docs.pylonsproject.org/projects/pylons-webframework/en/latest/advanced_pylons/entry_points_and_plugins.html) because it's dead-simple, well-tested, and prevents you from having to ever maintain anything.  It was so simple that I found it easier to generate an implementation than to talk to heavily about it.

It's 41 lines including an example plugin.  It took 4 hours and that was mostly refactoring the main function so that it had clearly separated stages (which didn't have to be done, but helped me untangle the way that control flows through the argument parsing).  It can very easily be the gateway to `pdoc` more than fulfilling its mission of replacing `epydoc`--because it finally makes your sanely generated Python docs available anywhere that people want to display them.

While you could port your HTML, HTTP, and Text implementations to be plugins, my implementation shows how they can be trivially attached as an afterthought and be perfectly functional.  Please check it out at [jvantuyl/pdoc@plugin_example](https://github.com/jvantuyl/pdoc/tree/plugin_example).

BurntSushi:
@jvantuyl I think we're talking past each other at this point. I'm happy to engage you, but I don't know how to at this point. I guess my last two suggestions are:

1. Forget about a "plugin system" and instead focus on what specifically you need to do. For example, if you need documentation in multiple formats, then a plugin system is not the only way to achieve it. (Neither is implementing each format in `pdoc` proper.)
2. Tone it down.

kstrauser:
@Friz-zy Ooh, very nice! I'm going to try using this as a skeleton to generate Swagger specs, too, so we can feed output directly into Swagger and get a nice API doc without parsing HTML and rewriting it. Thank you!

jvantuyl:
I think @kstrauser just hit the same issue.  We find the doc extraction part of `pdoc` very useful but the formatting part less so.

In the not too distant future, I want to do exactly what he wants to do but with Sphinx module / class / function directives. Cross references, too. (Markdown was always a compromise for me, but I went with it because it should be easy.)

I think we all agree that building that into `pdoc` would be a lot of work. What I'm less clear on is whether `pdoc` has a future as a solution to extracting docs for integration into other platforms or if it won't develop the capability of richly exposing what it's extracted.

BurntSushi:
@jvantuyl I'm interested in helping you, but I don't know how because you haven't actually told me, in detail, what problem you're trying to solve. :-(

jvantuyl:
I would like to do something like:

* `pdoc --swagger www/swagger.json --sphinx --sphinx-dir doc/api/ myapp`

Markdown was always a primitive step on the path to that goal, for me.

I would begrudgingly settle for:

* `pdoc --json --json-dir ./build/myapp.json myapp`

To be consumed by something like:

* `pdoc-to-sphinx ./build/myapp/myapp.json doc/api/`
* `pdoc-to-swagger ./build/myapp/myapp.json www/swagger.json`

BurntSushi:
I still don't understand the problem you're trying to solve, but it seems clear that we've hit an impasse.

I would like to try and get Markdown support added. Perhaps you will figure out a way to convert Markdown files to your preferred format.

In the future, I would consider adding a way to export documentation information to JSON.

jvantuyl:
The problem I'm trying to solve is integrating `pdoc` into a doc pipeline for, like, any project ever.

Could you point me to an example of the level of specificity you'd like? Because, short of giving you example output files, it doesn't get any more explicit than a command line; and this use case is, well, obvious.

BurntSushi:
Condescension is not welcome on this issue tracker. Please stop it. Responding with "it's obvious" to someone who is in good faith asking you to clarify is not constructive and it won't get us anywhere. I've in general found your language to be hyperbolic and exaggerated, and it makes it very hard to work with you.

---

I would like to know the details of the problem you care about. A "doc pipeline for any project, ever" is not meaningful to me. I (and plenty of others) use `pdoc` to provide documentation for many projects, but I've never needed any of the things you've asked for. There is nothing obvious to me about what you're asking for. I don't understand it.

An example specification might be:

1. Assume you have documentation in Markdown format.
2. Describe the sequence of steps to transform your Markdown documentation into formats you care about. Such a description probably includes shell commands, and perhaps even a short description of how the new format is consumed. (For example, I don't know what mkdocs is and I don't know what Swagger is.)

Notably, the specification should *never* mention `pdoc` at all. `pdoc` is a *solution* to the problem you care about. (Really, *any* documentation generator is a possible solution.) Therefore, incorporating it into your problem description makes it very hard for me to understand what it is you actually want.

An alternative specification would proceed similarly as above, but use JSON instead. I suspect this specification would take more effort to describe, but perhaps we could paper over the details of the JSON schema for now.

BurntSushi:
> Describe the sequence of steps to transform your Markdown documentation into formats you care about. Such a description probably includes shell commands, and perhaps even a short description of how the new format is consumed. (For example, I don't know what mkdocs is and I don't know what Swagger is.)

Building on this, you might also describe the shortcomings of starting with Markdown. What would it need to completely fulfill your use case?

jvantuyl:
I'll try to get you a response to this today (as I'm on a plane right now). I may open another issue for related concerns that aren't markdown specific. Markdown was always "step 1" for where I wanted to go and I don't think it's helpful to expand the scope of this issue too much.

jvantuyl:
*Disclaimer:  I think I've still fallen short of a specification here.  For the purposes of this issue, #100 is sufficient.  For my requirements (below), I think I should open new issues if we want to hash out a specification.*

Okay, here is what I hope is a more rigorous treatment of the requirements driving me here.  I'm numbering these to be able to refer to them later.  Otherwise the numbers have no significance.

1.  I would like to be able to produce handwritten documentation in a human-friendly format (i.e. user's guide, quickstart, etc.).
2.  I would like to be able to generate a static site for deploy.
3.  As part of that documentation, would like to have API documentation.
4.  I would like that API documentation to be autogenerated from the source code.
5.  I would like that API documentation to be styled with the same stylesheet as the rest of my documentation.
6.  I would like to generate versions of the documentation for hardcopy in addition to online versions.
7.  I would like the online documentation to be searchable.
8.  I would like seamless and unified navigation.
9.  I would like cross-references in the documentation.
10.  I would like the documentation generation pipeline to be robust.

Given those requirements, there are a few tools that fulfill parts of them, but not all.  I want to run through them because any specification is going to involve a pipeline along with some of these tools.  Taking the list requirement by requirement:

1.  [Github Pages](https://help.github.com/articles/creating-pages-with-the-automatic-generator/) (Markdown), Pelican (with ReST or Markdown), [Sphinx](http://www.sphinx-doc.org/) (with ReST), [mkdocs](http://www.mkdocs.org/) (with Markdown), etc.
2.  Jekyll (Markdown), Pelican (with ReST or Markdown), Sphinx (with ReST), mkdocs (with Markdown), etc.
3.  Sphinx (barely and painfully, with [sphinx-apidoc](http://www.sphinx-doc.org/en/stable/man/sphinx-apidoc.html) and [sphinx.ext.autodoc](http://www.sphinx-doc.org/en/stable/ext/autodoc.html#module-sphinx.ext.autodoc)).
4.  Sphinx (same as 3), [epydoc](http://epydoc.sourceforge.net/), [pydoc](https://docs.python.org/3.5/library/pydoc.html), `pdoc`
5.  Nothing seems to do this out of the box.  This is sort of achievable with a markdown-integrated pipeline; but not richly (i.e. not easy to mark, say, just functions with some markup).  When it is achievable, it's often by using other problematic constructs like embedding raw HTML.
6.  Sphinx and mkdocs do this.  Most others don't.
7.  Sphinx and mkdocs do this, but not for API docs.  Often this conflicts with static generation (req 2), though lunr.js is creating movement here.
8.  Sphinx and mkdocs do this, but not for API docs.
9.  Sphinx does this in a limited manner.
10.  Some of the above solutions don't meet this requirement when they start doing things like find/replace and other text processing on the output of previous steps.

These are a lot of options I could build such a specification around.  `pdoc` is the best tool that I've found for req 4.  The issue comes with handling its output.  Most of the "just output markdown from `pdoc`" options fall short because I don't get 5 (sometimes), 8, 9, or 10.  The existing HTML doesn't really count as robust either as parsing the output of the HTML template is dicey and, I feel, prone to be fragile.

The shortest-path-to-something usable option would be Sphinx-style ReST support.  The [Python Domain](http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain) and [cross references](http://www.sphinx-doc.org/en/stable/domains.html#cross-referencing-python-objects) make for beautiful documentation... with some caveats.  

If you go pure Sphinx, this is possible today (though only questionably fulfilling req 10) with some serious caveats.  Their solution is the pairing of `sphinx-apidoc` and `sphinx.ext.autodoc`.  `sphinx-apidoc` runs as a separate step, creates a tree of files filled with stubs (which are essentially disposable) and scans directories (not imports).  `sphinx.ext.autodoc` actually imports the modules in the interpreter running Sphinx.  This creates all manner of tedious tinkering necessary to handle projects in multiple versions of Python, makes it a pain to document standalone scripts, and (most limitingly) only works with Sphinx.  Also, entirely subjectively, I've had much better luck getting the right results from `pdoc` than from these tools.

With that in mind, I started in this conversation with mkdocs.  With it, you currently get everything except for the API documentation generation itself.  Notably, you get markdown which is much nicer to work with than Sphinx-encrusted ReST, limited navigation, and search.  That said, you lose cross-references, styling is basic,  and getting useful navigation can be tricky (because you've just got the markdown headers to work with).

So, right now none of the ways `pdoc` can give me documentation to have feature parity with the HTML or with the brittle Sphinx API stuff.  Certainly none of the tools seem to fit together with a strategy for cross-referencing or styling integration.  No solution is going to be perfect, but I think that level of integration is possible.

I've validated that #100 reaches what I'm looking for in basic integration (defined later), so I think this issue is done with the merge of the other (more current) commits.  For 2 and 3, I doubt any of this quite reaches the level of specificity to qualify as a specification.  I can take a stab at those and create separate issues for those if you think these requirements are worthwhile.

In terms of a roadmap, I imagine three tiers of integration being possible.  Right now, most integrations are basic or intermediate.  Even then, they don't very easily realize navigation and cross references.  Client-side search can also be elusive with some of them (epydoc is frustrating there).

1.  Basic: Very primitive markdown for inclusion in something that isn't going to have a rich import of docs.  #100 does this well.
2.  Intermediate:  Probably uses embedded HTML to get rich formatting of functions, classes, etc.  Could use anchors for in-file references but I wouldn't expect anything more.  #100 could perhaps be adapted to give this with more nuanced HTML generation (as seen in the current HTML output).  I'd expect this to be an entirely new pull request.
3.  Rich Integration:  In the long term, I'd like to extract the data to generate cross-references and richly style the elements.  The two targets that come to mind are HTML fragments and a Sphinx integration.  Both seem pretty achievable (I think).
  * HTML fragments would be easier to just drop into another template than the current fully-formed HTML pages (i.e. sans boilerplate CSS and some wrapper stuff).  That said, I don't know how to do cross-references with HTML fragments, so something magic would need to happen there.
  * Sphinx output would probably do the same job as `sphinx.ext.autodoc` but better (i.e. output the Sphinx ReST Python Domain format).  I'm also curious if some language features wouldn't require some retooling to style and cross-ref correctly (e.g. type annotations being the big one).
I'd also expect these to be entirely new pull requests, too.

I don't particularly want to start another argument over a plugin architecture, but I hope that these ambitions make clear why I was asking for one.  It seemed to be the lesser evil to bloating `pdoc` with with custom integrations (as internal web server, HTML, Markdown, Plain Text, and Sphinx would already be getting to be a pretty long list) or trying to generate some "best" neutral format (because you've already got a great set of classes that could be used directly).

cortesi:

I generally support the direction of this PR - we also want to be able to take pdoc output and integrate it in some reasonable way with our docs. I'm not sure if markdown output is the best way to go for this. It might be that "naked" HTML that can be rendered into a frame might be better. Let's discuss.

PR #37, opened 2015-03-11
Documentation aware of decorators and annotations

Hibou57:
Using [inspect.unwrap](https://docs.python.org/3/library/inspect.html#inspect.unwrap), to document decorated functions, rather than the decorator's inner wrapper function, which is needed as `Function.params` uses [inspect.getfullargspec](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec), which unfortunately does not follow the `__wrapped__` attribute.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -1096,6 +1096,9 @@ class Function (Doc):
         """
         params_spec = ', '.join(self.params())
         result_spec = self.result()
+        # TODO: it actually makes the return annotation, be displayed
+        # at the end of the arguments list, inside the parenthesis. This
+        # should be fixed.
         return (
             params_spec
             if result_spec is None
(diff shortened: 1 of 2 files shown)


Hibou57:
Now include annotations in function documentation.

This would need to be tweaked to looks better: the annotations are represented using the built‑in `repr`, which is not always good at it (ex. when the annotation is a type, like `int`, it displays as `<class 'int'>` instead of just as `int`). There is no way to workaround this using a custom `__repr__` or `__str__` of a custom class, as the annotations are ordinary tuples. May be an hook formater function could be an option. I don't know if there is anything like plug-ins with *pdoc*; if there are, this may be an option.

Hibou57:
Added a `Function.clean_ann` function.

Hibou57:
Note: as in this patch, `Function.spec(self)` has to handled parenthesis around arguments, to properly display any return annotation *next to* the closing parenthesis, the file `html.mako` was also changed too in this patch: `show_module(module)` does not add the parentheses any more, as they are now added by `Function.spec(self)`.

pdoc/__init__.py
```diff
@@ -1049,7 +1049,7 @@ def __init__(self, name, module, func_obj, cls=None, method=False):
         """
         super(Function, self).__init__(name, module, inspect.getdoc(func_obj))
 
-        self.func = func_obj
+        self.func = inspect.unwrap(func_obj)
```
BurntSushi: According to the docs for `inspect.unwrap`, it is new in Python 3.4. `pdoc` *must* remain compatible with at least Python 2.7, so you'll need to make this backwards compatible before I can merge it.

cortesi:

PR #38, opened 2015-03-12
Fix for `xrange` and Python 3

Hibou57:
`xrange` does not exist in Python 3, which have `range` only. Python 3 removed `range` from Python 2 and renamed `xrange` from Python 2 into `range`. See [What’s New In Python 3.0 / Views And Iterators Instead Of Lists](https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists).


BurntSushi:
Thanks!

PR #40, opened 2015-03-12
Adding a local modules list

Hibou57:
It can be hard to deal with a big list of modules in HTTP server mode, so I propose to add a local module list, based on the current working directory (the one from within which the server was started) and to add a link “Local packages” along with the already existing “All packages”. The links are also added in the index page, where there was none before.

### DIFF
--- a/pdoc/templates/css.mako
+++ b/pdoc/templates/css.mako
@@ -28,6 +28,9 @@
     font-size: 130%;
     margin: 0 0 15px 0;
   }
+  #nav > * + * {
+    margin-left: 1em;
+  }
 
   #top {
     display: block;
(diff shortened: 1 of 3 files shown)


BurntSushi:
I'm not sure if this fits well with the rest of `pdoc`. Here are my concerns:

1. My intention was for folks to do this by setting `PYTHONPATH` and using the `--only-pypath` option.
2. I'd prefer if there was exactly one list of modules served by any particular instance of `pdoc`. If you want multiple lists simultaneously, then I'd recommend multiple instances of `pdoc`.

I do actually like the idea of "serve docs for modules in current directory." It's much easier than munging with `PYTHONPATH`. If you change this to a flag, e.g., `--only-dir`, and have that override the main package listing with modules from the current directory, then I think I'd accept that.

cortesi:

PR #42, opened 2015-03-13
Avoid trying to import from non-existant packages

Hibou57:
Without this patch *pdoc* tries to import package's module as if there were submodules of modules as packages, for each package's modules.

Ex. if a package `P` contains modules `M1`, `M2`, `M3`, *pdoc* tries to import `P.M1.M1`, `P.M1.M2`, `P.M1.M3`, `P.M2.M1`, … and so on. This patch fix this, which I believe is an issue.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -594,21 +594,23 @@ class Module (Doc):
         else:
             pkgdir = getattr(self.module, '__path__',
                              [path.dirname(self.module.__file__)])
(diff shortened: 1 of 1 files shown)


Hibou57:
Please, note although this patch is reported as changing 17 lines, it's actually a lot simpler: it only adds one line, `if self.is_package():` and indents what's beneath.

BurntSushi:
Thanks!

PR #44, opened 2015-03-15
Avoid filtering out variables on their value

Hibou57:
Without this patch, *pdoc* will filter-out variables whose values are an instance of a class defined in an imported package.

How to solve: remember *explicitly declared variables* by remembering the keys of `vardocs`, then in `__is_exported`, don't filter-out what may seems to be from another module and ends to be a variable whose value is an instance of a class declared in another module.

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -559,6 +559,7 @@ class Module (Doc):
             vardocs = _var_docstrings(tree, self, cls=None)
         except:
             pass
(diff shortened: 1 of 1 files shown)


Hibou57:
Please, note although this patch is reported as changing 17 lines, it's actually a lot simpler: it only adds one line, `if self.is_package():` and indents what's beneath.

BurntSushi:
I buy it. Thanks!

PR #45, opened 2015-03-17
Fix wrong HTML body height

Hibou57:
The `height: 100%;` is erroneous, as it gives the body an height which is 100% of that of the view-port, which is not relevant for a content which is to be scrollable. To see it, one may give the body a non-white background color, say orange (as an example), then scroll the content: the top has an orange and the rest of the scrollable content has a white (default) background.

### DIFF
--- a/pdoc/templates/css.mako
+++ b/pdoc/templates/css.mako
@@ -2,6 +2,7 @@
   html, body {
     margin: 0;
     padding: 0;
+    min-height: 100%;
   }
   body {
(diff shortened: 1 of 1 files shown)


somenxavier:
@Hibou57 I'm interested in your patch. Can you submit your pull request to [my fork](https://github.com/somenxavier/pdoc)? Thanks in advance.

BurntSushi:
Thanks!

PR #61, opened 2015-07-20
removing max-width

somenxavier:
In screens with big screen, there is no need for max-width. If you don't like that, perhaps we could differentiate among different types of screens....

### DIFF
--- a/pdoc/templates/css.mako
+++ b/pdoc/templates/css.mako
@@ -13,7 +13,6 @@
   }
   #content {
     width: 70%;
-    max-width: 850px;
     float: left;
     padding: 30px 60px;
     border-left: 1px solid #ddd;


somenxavier:
I apply to my release [0.3.1a](https://github.com/somenxavier/pdoc/releases/tag/0.3.1a)

BurntSushi:
No, indeed, I do not like this at all. I hate long lines and think they are hard to read. Even on big screens, the width should be capped.

somenxavier:
It think it's a personal aesthetic question for you. For me is a constraint when you put tables in documentation: you get rows split

BurntSushi:
I'm open to ways of addressing your specific problem (rows split in tables) without making text stretch the width of the browser window.

BurntSushi:
If we can't agree on something, you can certainly modify the existing templates to your taste. `pdoc` has a `--template-dir` flag for this reason.

somenxavier:
Perhaps a border=1 could be another thing to consider in this topic. Thanks for the references.

somenxavier:
Just a comment, instead of putting ` -    max-width: 850px; ` why not put max-width with relative units. It's common recognized that relative units are best than absolute ones in HTML. Something like `max-width: 70%`.

PR #69, opened 2015-10-12
Fixed --template-dir for http server.

pieterb:
Command line option `--template-dir` was not propagated to pdoc subprocesses, causing `pdoc` to fail if options `--http` and `--template-dir` were combined. This patch fixes that.


BurntSushi:
Great! Thanks! :-)

PR #70, opened 2015-10-21
fix Python 2 incompatibility

jvantuyl:
The fix introduced in dc57c4a1 doesn't work on Python 2 (because builtins isn't
always importable).  This commit attempts to shadow xrange with itself, and
falls back to range when it's not available (which accomplishes the same
thing).


BurntSushi:
Ah, much better! Thanks!

PR #72, opened 2015-10-21
ignore files generated by development installs

jvantuyl:
When installed with `pip install -e`, some files are generated.  These should
be ignored so that they aren't inadvertently checked in.


BurntSushi:
Great, thanks!

Issue #84, opened 2016-01-14
Html generation leaves out the first character from python imports


IAlwaysBeCoding:
So, I found an odd bug. Don't really know how to fix it , but I know how to get around it.

Here is a screenshot:
https://imgur.com/w6Z5Ro3

Example:

```
from collections import OrderedDict
```

Will produce source code without the first character, only if the first line is blank. To fix this, just make sure the first line isn't blank, and will produce the correct output.

Also, just wanted to say that your project is FREAKING amazing. I am thinking of using it for stitching together something simple that I can use to upload my documentation to gitbook.com . Sphinx is kinda too big, although is great it doesn't work well with Markdown. Plus, I saw that your project also processes any markdown too in doc strings which is super useful. Thanks for this god-send project, just what I was looking for .

BurntSushi:
That is a strange bug indeed! I haven't reproduced it yet, but I wonder, is the HTML correct and this is just a styling issue? Or is the first character actually dropped from the output altogether? (Either would be weird!)

IAlwaysBeCoding:
I am not sure, yet haven't taken a look at the code. Here is the html of that screenshot
https://gist.github.com/IAlwaysBeCoding/3192fe35a827646a2807.

And it is actually stripping the first character of each line, the html shows that it is being stripped and not just a styling issue.

Now , you made me curious let me take a look why that is happening and do a PR.

IAlwaysBeCoding:
Okay, found the bug. It's here https://github.com/BurntSushi/pdoc/blob/master/pdoc/templates/html.mako#L41

It basically tries to get the indent of the first line, but since the first line is an empty line the indent is 1 and not the expected 0. So, on line 41 basically says give me every single line from source code but starting from the `index`  whose value equals  `indent` .

I am not sure why it needs to get the indent though, so am afraid of removing that thinking it might mess something else. However, if you remove line 41 then it should work.

BurntSushi:
@IAlwaysBeCoding the reason why it's mucking with indents is if the code *is* indented, then it doesn't look so good to just put that in `<pre>` tags as-is (since the indent will show). In the case of showing the full module, mucking with the indent isn't needed.

I think I'd probably recommend fixing the indent code. That is, compute the indent based on the first non-empty line instead of the first line. (There may be other cases where this fails, but I'd be happy with that simple fix for now!)

IAlwaysBeCoding:
[...]

PR #86, opened 2016-01-14
Remove html stripping bug from html.mako

IAlwaysBeCoding:
Removes a bug where it strips the first character of every line in the source code if you leave first line blank.

### DIFF
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -38,7 +38,11 @@
     Returns one string with all of the source code.
     """
     base_indent = len(indent.match(lines[0]).group(0))
-    base_indent = 0 if lines[0].isspace() else base_indent
+    base_indent = 0
+    for line in lines:
+      if len(line.strip()) > 0:
+        base_indent = len(indent.match(lines[0]).group(0))
+        break
     lines = [line[base_indent:] for line in lines]
     if not use_pygments:  # :-(
       return '<pre><code>%s</code></pre>' % (''.join(lines))
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -38,6 +38,7 @@
     Returns one string with all of the source code.
     """
     base_indent = len(indent.match(lines[0]).group(0))
+    base_indent = 0 if lines[0].isspace() else base_indent
     lines = [line[base_indent:] for line in lines]
     if not use_pygments:  # :-(
       return '<pre><code>%s</code></pre>' % (''.join(lines))


IAlwaysBeCoding:
@BurntSushi  I like your solution better

IAlwaysBeCoding:
Well, am reopening this PR to fix #84 .

pdoc/templates/html.mako
```diff
@@ -38,6 +38,7 @@
     Returns one string with all of the source code.
     """
     base_indent = len(indent.match(lines[0]).group(0))
+    base_indent = 0 if lines[0].isspace() else base_indent
```
BurntSushi: Could you change this to:

```python
base_indent = 0
for line in line:
    if len(line.strip()) > 0:
        base_indent = len(indent.match(lines[0]).group(0))
        break
```

(Assuming it works. :-))

IAlwaysBeCoding:
Okay, got it its fixed. You just had a small spelling mistake,  in the `for line in line`. Other than that it works.

BurntSushi:
Yay! @IAlwaysBeCoding thank you! And thanks for your patience. :-)

Enjoy!

IAlwaysBeCoding:
Oh, no ! Thank you, am kinda new to this thing of doing PR's . Thanks for letting me do a PR. I liked that little interaction we had there :+1: . Also, am curious since you seem to write a lot in the Rust language, how would you do a quick comparison between pdoc(python) and Docopt(rust) . Eventually, am going to read that one next.As for my new years resolution,I am trying start another big journey like I did with python a few years ago(focus on one language as my primary language), and I wanted to get an opinion of someone who came from a python background and picked up Rust extensively.

BurntSushi:
@IAlwaysBeCoding I <3 Rust. (I'm a [member of the library team](https://www.rust-lang.org/team.html#Library).) Docopt(rust) is actually a command line argument parser in the style of [docopt.org](http://docopt.org/) (whose reference implementation is actually written in Python). So the two projects aren't particularly comparable.

With that said, Rust is a great language to learn. If you don't have a systems language background (e.g., C, C++), then it might be a little intimidating at first, but there are plenty of folks learning Rust without systems language experience. Plus, we're all very friendly with beginners. :-)

Generally speaking, the class of problems one might want to solve with Python as opposed to Rust is going to be very different. (Not to say there isn't any overlap though!) Rust is going to give you a static type system, which means you'll need to get your code past the compiler before it can run. Rust also doesn't have a garbage collector, so you'll need to reason about the lifetime of data in your program. It can be a little daunting at first, but the [Rust book](https://doc.rust-lang.org/stable/book/) is a great place to start.

IRC is a wonderful resource. I'm usually around when I'm not at work. :-)

IAlwaysBeCoding:
@BurntSushi awesome! thanks for the reply. I will take a look at the rust IRC.I really like their cargo system they have, much better than python pip if you ask me. I hardly install anything with direct pip from the main repository. I always install from github(on the condition there is a setup.py there) as it has proven much better for stability.

Issue #35, opened 2015-02-27
Parsing Epytext


biomadeira:
Sorry the ignorance. Is there a way of 'forcing' pdoc to parse docstrings in Epytext format as the example below:

``` python
def load_config(filename, option):
    """
    Loads and tests input parameters.

    @param filename: input filename
    @param option: option name
    @return: returns a valid config value
    """
```

Thanks

BurntSushi:
No, but you could modify the templates to do whatever you want pretty easily: https://github.com/BurntSushi/pdoc/blob/master/pdoc/templates/html.mako#L66

biomadeira:
Nice. Thanks!

johnyf:
I am also interested in this. Being an `epydoc`user myself, for some time now I have been using Epytext with markdown formatting. Though I am not aware of a suitable doc tool, I was planning to write one. The closest thing is to adapt `pdoc` for recognizing `@param, @type, @return, @rtype`, and also link automatically `Name`, replacing the Epytext `[Name]`, if `Name` is found to be an existing class/function/etc somewhere in the package.

peterjc:
I missed this issue when I filed #111 for the wider issue of supporting some of the markup languages as supported in epydoc (which does not support markdown).

cortesi:
Please see #153 for a unified discussion ticket about docstring flavors.

PR #101, opened 2016-04-07
Add support for Google Style

j6k4m8:
[Google docstrings](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html) are increasingly common. There are definitely [other users](https://github.com/BurntSushi/pdoc/issues/89) who use Google-style strings as well.

I've implemented some very basic parsing for Google-style docstrings in `html-google.mako`. Seems like a project like this could benefit from natively supporting a few common docstring styles, so I'd be willing to generalize this to numpy-style as well and add a flag `--style=google` perhaps.

Looking forward to hearing your thoughts! cc @BurntSushi for discussion as to if we should merge, cc @igordertigor because I thought you might be interested in seeing this. cc @willgray in case you're interested.
## Google Style

```
"""
A short description of the function goes here.

Arguments:
    argname (type): Description.
    secondarg (type : default): Description.

Returns:
    boolean: True on success.
"""
```


j6k4m8:
Take a look [here](https://github.com/BurntSushi/pdoc/pull/101/files#diff-e5116ef71c494d0e79d056e48749ba23R72) for the actual change. (I moved it to another file to preserve existing behavior.) Essentially just two very dumb regexes.

BurntSushi:
@j6k4m8 Thanks! Two points:

1. A duplicate of `html.mako` with just a few changes is not something I'm personally interested in maintaining.
2. It seems like this doesn't need to be in `pdoc` proper, and someone could instead just modify their own `html.mako` template?

Popping up a level, if we want to support multiple doc string formats, then it might indeed be better to add a CLI flag with a list of supported formats. I am slightly worried about the maintenance burden, however.

j6k4m8:
That's a great point — one of the reasons I added the `don't-merge` was so I could get your input on how best to extend pdoc — I'm pretty unfamiliar with mako. 

Would you be amenable to a variety of parser functions that can be triggered via CLI flags? Looking back at this, I'm not sure why I implemented this way...

BurntSushi:
@j6k4m8 I personally would prefer encouraging one format instead of a divergence of formats. I think the specific thing I'm worried about is supporting the currently very basic and simple linking mechanism. It's a Markdown specific regex that may not apply to other doc string formats. In particular, reST is quite a bit more sophisticated and I'm not particularly interested in doing the work for full integration with it.

If we can find a way to make this really simple, then I could be persuaded to get on board.

j6k4m8:
...alright cleaned it up a ton. You can now call `mark` with a `preprocessor=Preprocessors.GOOGLE`, for instance — and before the text is passed to the `markdown` parser, it's processed via an optional function. 

The output of the preprocessor is still markdown — I'm very much opposed to adding the extra complexity of supporting reST (which sucks anyway) or other formats beyond simply supporting common docstring formats that are nearly markdown anyway.

pdoc/templates/html.mako
```diff
@@ -59,15 +59,35 @@
...
     if not module_list:
       s, _ = re.subn('`[^`]+`', linkify, s)
 
     extensions = []
+
     if use_pygments:
-      extensions = ['markdown.extensions.codehilite(linenums=False)']
+      extensions.append('markdown.extensions.codehilite(linenums=False)')
+
+    if preprocessor:
```
BurntSushi: Where is this set?

scripts/pdoc
```diff
@@ -77,6 +77,9 @@ aa('--http-dir', type=str, default=default_http_dir,
         'server.')
 aa('--http-host', type=str, default='localhost',
    help='The host on which to run the HTTP server.')
+aa('--docstring-style', type=str, default=None,
```
j6k4m8: Pass `--docstring-style=google` to parse with a Googly flavor.

j6k4m8:
Run this:
```
pdoc ndio --html --html-dir="pdocs" --overwrite --docstring-style=google
```

...to get this:
image

j6k4m8:
@BurntSushi sorry for the long, drawn out PR comments-section! But I think this is working now — should be pretty clean code, and as far as I could tell, it doesn't get in the way if you want to just run without a `docstring-style` flag. Looking forward to hearing your thoughts!

realitix:
Nice j6k4m8 !
Google documentation is very clean and elegant.

igordertigor:
Cool!

j6k4m8:
Glad you guys are excited about this too!! I haven't looked at this in a while, so it's possible it's not the best approach anymore. But I'll take another look this week

peterjc:
How about calling the new command line argument ``--docformat`` to match ``epydoc``?

Do you think this could be adapted to insert an RST to markdown preprocessor (or other options like epytext), referencing #35 and #111

j6k4m8:
I'm amenable. But ultimately up to @BurntSushi!

I tried to write it as modularly as possible so that you can drop in whatever flavor of parser you like

benjamingorman:
Any idea of if/when this PR will be accepted?

cortesi:

While I'd like to have as few fiddleable knobs in pdoc as posssible, I think having support for some (lightweight and well-defined) alternative comment styles might be plausible. We've decided that complex formats like RST are not on the roadmap.

cortesi:
Please see #153 for a unified discussion ticket about docstring flavors.

PR #122, opened 2017-01-24
Properly handling non-ascii characters in desc

mbeloshitsky:


BurntSushi:
I'm not sure this is the right place. Ideally, the `docstring` itself would be a Unicode string.

PR #128, opened 2017-04-20
Mark asynchronous functions as `async def`

kchmck:
This change marks [asynchronous/coroutine functions](https://docs.python.org/3/reference/compound_stmts.html#async-def) as `async def` in the generated docs, since they're used and overridden differently than regular functions.

It should be backwards compatible with python < 3.5 as well.


mhils:
Thanks! 🍰

PR #129, opened 2017-04-20
Use Python 3 highlighter when ran in Python 3 environment

kchmck:
Among other things, this adds highlighting for the `async` keyword

The py3 lexer was added in [version 0.10](http://pygments.org/docs/lexers/#pygments.lexers.python.Python3Lexer)


mhils:
Thanks! 🍰

PR #133, opened 2017-05-30
Fix inheritance labeling in Python 3

### LINKED ISSUE #132: Docstring inheritance fails under Python 3
When `pdoc` (v0.3.2, running under Python 3) is run on the following minimal example:

```python
class A:
    x = 1
    """One class var"""

    y = 2
    """Another class var"""

class B(A):
    x = 5
    y = 10
```

it fails to correctly register the inherited docstrings for class `B`.

In particular, exactly one of `B.x` or `B.y` (either has been observed) will have its docstring correctly inherited from the corresponding member of `A`, and the other will not.

kchmck:
In Py3, the filter() call creates an iterator which is exhausted by the
second call to search(). This was causing some methods to
nondeterministically not receive the inheritance annotation in the
generated docs.

Seems to fix #132

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -992,8 +992,7 @@ class Class (Doc):
         variables are only discoverable by traversing the abstract
         syntax tree.
         """
-        mro = filter(lambda c: c != self and isinstance(c, Class),
-                     self.module.mro(self))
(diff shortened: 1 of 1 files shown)


BurntSushi:
Awesome, thanks!

Issue #111, opened 2016-07-29
 __docformat__ reStructuredText/rst etc input as well as Markdown


peterjc:
According to the README,

> It is intended that pdoc will be a replacement for the unmaintained epydoc.

epydoc defaults to its own simple epytext markup but supports additional markup languages, reStructuredText, Javadoc and plaintext:

http://epydoc.sourceforge.net/epydoc.html#alternate-markup-languages

They specifically recommend using reStructuredText if epytext is not expressive enough:

> Epytext is intentionally very lightweight. If you wish to use a more expressive markup language, I recommend reStructuredText.

http://epydoc.sourceforge.net/epydoc.html#the-epytext-markup-language

I am surprised to see that currently pdoc appears to assume the docstrings are written using Markdown notation (which is _not_ supported by epydoc).

Would you consider supporting other markup languages (I would particularly like to use reStructuredText aka rst) as indicated by the `__docformat__` attribute support and/or a command line switch as in epydoc?

e.g.

``` python
__docformat__ = "restructuredtext en"
```

or,

``` python
__docformat__ = "plaintext en"
```

or

``` python
__docformat__ = "epytext"
```

automaticgiant:
I'd support this solely based on the ability to then use `pdoc --http` to view the 'canonical' doc format (that of the standard library, and whatever copies its format of choice).

cortesi:
Hi there! We've decided that support for markup format beyond markdown is not a goal at this point. We feel this is the best way for pdoc to remain a small, clear, predictable tool.

cortesi:
Please see #153 for a unified discussion ticket about docstring flavors.

Issue #63, opened 2015-08-26
Doctests are interpreted by the HTML template


mikael-s:
When producing the output in HTML mode, doctests are simply interpreted as if they were part of markdown syntax and are replaced by (three) blockquotes.

For instance if you have the following code in a `test.py` file:

``` python
def f():
    '''
    >>> f()
    2
    '''
    return 2
```

Launching `pdoc` on that file with `pdoc --html test.py` will produce a HTML file where the docstring is replaced by the following code:

``` html
<blockquote>
<blockquote>
<blockquote>
<p>f()
2</p>
</blockquote>
</blockquote>
</blockquote>
```

Therefore the `>>>` are removed and the line return is ignored which makes the doctests hardly readable.

peterjc:
I would retitle this slightly, but can confirm that pdoc does not nicely render docstests in docstrings. Example:

```python
from __future__ import print_function

def double(x):
   """Return double the input value.

   >>> double(5)
   10
   >>> double(1.25)
   2.5

   """
   return x * 2

if __name__ == "__main__":
    for v in [0.1, 3, 123]:
        print("Doubling %r gives %r" % (v, double(v)))
```

Then

```
$ pdoc --version
0.3.2
$ pdoc --html example.py 
$ open example.m.html
```

Gives:

```html
...
  <section id="section-items">

    <h2 class="section-title" id="header-functions">Functions</h2>
      
  <div class="item">
    <div class="name def" id="example.double">
    <p>def <span class="ident">double</span>(</p><p>x)</p>
    </div>
    

    
  
    <div class="desc"><p>Return double the input value.</p>
<blockquote>
<blockquote>
<blockquote>
<p>double(5)
10
double(1.25)
2.5</p>
</blockquote>
</blockquote>
</blockquote></div>
  <div class="source_cont">
...
```

<img width="856" alt="screenshot 2017-09-08 10 48 47" src="https://user-images.githubusercontent.com/63959/30206213-4861f334-9483-11e7-8a22-361a0dc8aa63.png">

peterjc:
This seems to be a limitation of the markdown rendering engine - which makes sense after all, ``>>>`` is not part of the Markdown language.

If pdoc supported reStructuredText as input, that does include support for ``>>>`` doctests - see #111.

cortesi:
Please see #153 for a unified discussion ticket about docstring flavors.

PR #137, opened 2017-09-08
Add --docformat argument for reStructuredText or plaintext input

peterjc:
This leaves markdown as the default for rendering the docstrings, but adds a new ``--docformat`` option (named after the ``epydoc`` option) allowing the use of reStructuredText (via ``docutils``) or plain text (using an HTML ``<pre>`` block).

This will be useful for anyone hoping to migrate from epydoc which did not support Markdown, but did support its own format epydoc and the Python standard reStructuredText.

This will also indirectly greatly improve the display of doctests or Python snippets using the `>>>` notation which the Markdown rendering does not understand (see #63).

It would be relatively simple to make the ``markdown``/``docutils`` import lazy, so that people using one input format would not require the other dependency be installed.

This would close the main goal of #111 (although it does not make life complicated by looking at the per-file settings in any ``__docformat__`` declaration).

Thank you to @j6k4m8 whose work on  #101 was very useful. I would be happy to include his work by adding GitHub Flavoured Markdown as another option here, e.g. ``--docformat gfm``?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ Prominent features include:
   [namedtuple](http://docs.python.org/2.7/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields)),
   the special variable `__pdoc__` can be used in your module to
   document any identifier in your public interface.
-* Usage is simple. Just write your documentation as Markdown. There are no 
-  added special syntax rules.
+* Usage is simple. Just write all your docstring documentation as Markdown
+  (default), or reStructuredText. There are no added special syntax rules.
 * `pdoc` respects your `__all__` variable when present.
 * `pdoc` will automatically link identifiers in your docstrings to its
   corresponding documentation.
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -238,7 +238,8 @@ object's `directories` attribute.
 
 
 def html(module_name, docfilter=None, allsubmodules=False,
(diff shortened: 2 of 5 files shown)


j6k4m8:
Awesome!!

I really like this idea. Feel free to PR this on my fork too if you'd like; would love to merge this in right away and get to use it!

peterjc:
I think it might need some merge conflicts resolved, but I have no objection to you doing this @j6k4m8

skjerns:
Is there any update on the RST support?

skjerns:
Thank's I'll have a look into `sphinx-apidoc`. 
If you have any hint on how to easily create docs from reST docstrings, let me know.

cortesi:
Hi there! We've decided not to support RST in pdoc. I realise this might be a painful choice, but we feel pdoc will be better if it remains small and focused, with a single markup format.

karolzmijewski:
Is it finally supported or not? I'm a bit confused, because regarding [doc](https://pdoc.dev/docs/pdoc/render.html) we can choose `restructuredtext` as a docformat:
`docformat: Literal['markdown', 'google', 'numpy', 'restructuredtext'] = 'restructuredtext',`
but when I try to use it, It doesn't parse the docstring as I expect

peterjc:
@karolzmijewski this pull request was not merged, see discussion on #153 instead.

Issue #123, opened 2017-02-08
Unable to run pdoc from command line


CadBrad:
This question was closed in the past, but the solution didn't solve it for me. I'm running Windows 7 and python 2.7

I'm installing pdoc globally using "pip install pdoc".  It says that it pip installs correctly, puts a file called 'pdoc' (no extension) into my Python27/Scripts (which is on my PATH), however when I type "pdoc" in the command line I get the following...
'pdoc' is not recognized as an internal or external command, operable program or batch file'

The previous closed issue was solved by running "pip install --system pdoc" but that was for Ubuntu and doesn't seem to be an option for Windows.

BurntSushi:
This isn't a problem with `pdoc`. You need to make sure `Python27/Scripts` is in your `%PATH%` environment variable. Alternatively, write out the full path: `C:\foo\bar\Python27\Scripts\pdoc.exe`.

CadBrad:
As I originally mentioned, Python27/Scripts is in my PATH.

You mention the full path ending in "pdoc.exe", however my pdoc doesn't have any extension and is just shown as "pdoc".  Should pdoc have the extension .exe?

BurntSushi:
I guess it shouldn't have the `.exe` extension.

Honestly, I don't know. I don't use Windows, but given the error message, it seems like a Windows specific problem. I would suggest Googling around for how to install Python programs on Windows.

robomojo:
the pdoc file is a python file. You should use it by running it with python and passing arguments.
```
C:\Python27\python.exe C:\Python27\Scripts\pdoc my_module.py
```

CadBrad:
When I run `python pdoc my_module.py` I get the error "python: can't open file 'pdoc': [Errno2] No such file or directory".  Again, I confirmed pdoc is on my PATH.

Please bear in mind I've only tried doing `pip install pdoc`, and haven't tried to build from the source.  I found another solution that was faster for my situation and didn't need to move forward with anymore troubleshooting.

I appreciate the software being opensource, and I hope by posting my installation issues it can help future Windows users who may experience this as well as the developers who may not have access to Windows environment for testing.

BurntSushi:
@CadBrad Another possibility is `python -m pdoc my_module.py`?

CadBrad:
That returns the following:
"C:\Python27\python.exe: No module named pdoc.__main__; 'pdoc' is a package and cannot be directly executed"

BurntSushi:
[...]

Issue #52, opened 2015-05-29
Pdoc commandline script name conflict


karimbahgat:
When using pdoc for the first time I encountered an error with the commandline script that got placed into Python27/Scripts. First off, the file did not have the .py extension, so I had to add it. Secondly, when the commandline script tried importing pdoc from site-packages it just ended up importing itself (because the commandline script had the same name as the site-packages pdoc), resulting in a nameerror when trying to run the `import_module()` function. Changing the name of the commandline script fixed this error. 

I am surprised noone else have gotten this error? Should be a quick fix anyway. I am on Windows 8, 32 bit Py27.

marcos-alfredo:
Had the same issue where after installing pdoc, it created a script without extension.
Using Python 2.7 and Windows 7.

cortesi:
Pdoc now targets Python 3.6 and later. Please let us know if you still see this issue there. Thanks!

Issue #20, opened 2014-12-09
Windows Powershell


suidobashi:
Installed using pip to a Windows 8.1 Powershell environment.

Was unable to run the pdoc script as a shebang script. (Powershell requires a filename extension to make the script executable.)

Workaround was to run directly from the python interpreter.

BurntSushi:
Does anyone know how to fix this? I don't use Windows. What if I use a CLI hook in `setup.py`?

mhils:
Thanks for reporting and sorry for the slow turnaround time. This should be fixed with the latest release. 😃

PR #138, opened 2017-11-06
Fix windows install

### LINKED ISSUE #127: After installing pdoc with pip it not recognised as an executable on Windows
Tried navigating to \Scripts, same result.

sjpb:
This:
- Fixes Windows installs, including anaconda (it's not special in this regards). Having reviewed the issues, I believe this will fix #127, #123, #52 and #20.
- Should not affect Linux installs (but I can't test this unfortunately).

As in #127 you said you didn't use Python on Windows I'll try to explain:

Basically, the Windows install issue was down to using distutils' `scripts` argument in `setup.py` to create the CLI tool. This simply copies the named script into the Scripts folder on install. This didn't work on Windows as you ended up with a no-extension, non-executable file on your $PATH.

What this PR does instead is to use setuptools' `entry_points:console_scripts` feature, which is apparently now the preferred/cross-platform way to create CLI tools. Instead of just copying a file, it generates a wrapper for the appropriate platform (i.e. an .exe on Windows). However the wrapper has to point to an entry point inside the package, hence the change from `scripts/pdoc` to `pdoc/cli.py` and the addition of `cli()`. I can't test but I suspect the "wrapper" on Linux will simply be the script itself, chmoded to executable.

Motivated to fix this as pdoc looks like exactly what Python needs - many thanks!

### DIFF
--- a/pdoc/cli.py
+++ b/pdoc/cli.py
@@ -389,7 +389,9 @@ def process_html_out(impath):
         print(out)
 
 
-if __name__ == '__main__':
+def cli():
+    """ Command-line entry point """
+
     if args.version:
         print(pdoc.__version__)
         sys.exit(0)
@@ -504,3 +506,6 @@ if __name__ == '__main__':
         quit_if_exists(module)
         html_out(module, args.html)
         sys.exit(0)
+
+if __name__ == '__main__':
+    cli()
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
                                 'UNLICENSE', 'INSTALL', 'CHANGELOG']),
                 ('share/doc/pdoc', ['doc/pdoc/index.html']),
                ],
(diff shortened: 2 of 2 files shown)


relu91:
I've tried to install pdoc using this changes. I confirm that it's working on a Windows 10 machine. 

Thanks.

mhils:
Thanks! 🍰

dhamacher:
I am not able to use pdoc on Windows 10 and pdoc is not created as cli app in my Scripts folder. I went through the steps listed in this issue to troubleshoot and see if there is another reason why pdoc is not recognized with no success, so I am asking if something changed recently? Thank you.

sjpb:
You'll need to raise an issue on the pdoc issue tracker - this PR has been merged and I don't know what's changed after this I'm afraid.

PR #139, opened 2018-01-22
Only show *relevant* ancestors by pruning self and object from MRO

kernc:
`cls` will always be in `cls`'s MRO, and so will `object` that everything inherits from. Hence, these two are redundant?

pdoc rocks; this is how Guido had imagined it! 😄

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -697,15 +697,16 @@ class Module (Doc):
 
     def mro(self, cls):
         """
-        Returns a method resolution list of documentation objects
+        Returns a method resolution list of ancestor documentation objects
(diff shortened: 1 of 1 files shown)


cortesi:
Hi there! I'm going to merge this and the other PR you've submitted - many thanks for your work.

I just want to note your avatar in passing. I have seen that you've made comments elsewhere that you don't intend for it to be offensive, and I'm aware of the use of this symbol in many eastern spiritual traditions. That said, you must be aware of the unpleasant connotations that this will have for people, and not everyone is going to dig in to see what the intent behind your use is. If you plan to contribute to further pdoc (and we hope you do), we'd much rather you use something less provocative.

PR #145, opened 2018-05-06
Fix mako default encoding from ascii to utf-8

kernc:
Fixes https://github.com/BurntSushi/pdoc/issues/134
Replaces/Closes https://github.com/BurntSushi/pdoc/pull/122

See also:
http://docs.makotemplates.org/en/latest/unicode.html#specifying-the-encoding-of-a-template-file

### DIFF
--- a/pdoc/templates/html.mako
+++ b/pdoc/templates/html.mako
@@ -1,3 +1,4 @@
+## -*- coding: utf-8 -*-
 <%
   import re
   import sys


mhils:
Thanks for the PR! 🍰 😃 

On another note: We know that swastikas are not necessarily a nazi symbol (quite the opposite sometimes), but we'd nonetheless appreciate if you could change your avatar for future contributions as it may reasonably be misinterpreted and offend people. I think we're all in favor of a friendly atmosphere here. :) Thanks for your understanding!

PR #162, opened 2018-08-15
Read __version__ from the correct file. Fixes #161

### LINKED ISSUE #161: Building results in version 0.0.0
I ran `python3 setup.py sdist` on the latest source and got a resulting `dist/pdoc-0.0.0.tar.gz` file.

This is because `setup.py` looks for `__version__` in `pdoc/__init__` but it's actually located in `pdoc/doc.py`.

`setup.py` should look in `pdoc/doc.py` instead, and `pdoc/__init__.py` should contain a `from .doc import __version__` so `import pdoc; pdoc.__version__` works as expected.

ecederstrand:


mhils:
Thanks for the bunch of of good PRs and sorry for the slow turnaround - we've all been on vacation or otherwise busy, should be better from now on. 

Looking at #163 and this one, how about we just move `__version__` to `__init__.py` directly? I think that was just inadvertently moved in https://github.com/mitmproxy/pdoc/commit/8b3a7f8bf5ec8b47eb57215b28c1221f435615ad.

ecederstrand:
I have no problems with that. I pushed a fix just now.

mhils:
Thanks!

PR #172, opened 2018-09-08
Define unicode in Python 3

cclauss:
Without this change __NameError__ could be raised at runtime because __unicode__ was removed in Python 3 because all __str__ are Unicode.

flake8 testing of https://github.com/mitmproxy/pdoc on Python 3.7.0

$ flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
```
            if isinstance(el, str) or isinstance(el, unicode):
                                                     ^
1     F821 undefined name 'unicode'
1
```

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -1,6 +1,10 @@
 import ast
 import inspect
 
+try:
+    unicode        # Python 2
+except NameError:
+    unicode = str  # Python 3
 
 __pdoc__ = {}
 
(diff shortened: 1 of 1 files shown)


mhils:
We don't support Python 2 anymore, so I guess the correct approach here is to just remove the unicode reference. :) What do you think?

mhils:
Thanks! 😃

PR #177, opened 2018-09-23
Fix hyperlinking and other HTML stuff

kernc:
The commits seem disjoint, but really are just a series of patches that pretty much restores (and a bit builds upon) the previous HTML output behavior.

Please see commits individually.

Changes of note:
* Markdown2 is back to Markdown1. Fenced code works. As does everything else. See the commit message.
* Links in the sidebar and links in markdown now work.
* Documenting only public submodules/members has been restored.
* Pygments has been replaced with client-side highlighting with Highlight.js.
* A couple other HTML template changes.

Of course, everything is open to discussion.


pdoc/doc.py
```diff
@@ -588,15 +608,18 @@ def __public_objs(self):
...
         _pdoc = getattr(self.module.module, "__pdoc__", {})
 
-        def forced_out(name):
+        def force_skipped(name):
             return _pdoc.get("%s.%s" % (self.name, name), False) is None
 
         def exported(name):
             exported = name == "__init__" or _is_exported(name)
-            return not forced_out(name) and exported
+            inheried = name not in self.cls.__dict__
```
kernc: This now skips listing inherited members that have not been overridden. E.g.

```py
class C:
   def f(self): pass

class D(C):
   def g(self): pass
```
Here for `D`, only `g()` will be documented, and for `C`, `f`.

The rationale is one can look into linked ancestors for the inherited members, less repetition etc.

Even though it looks quite pretty in the project I'm working with, I imagine not all projects would like/expect this. Can be made a setting if `__pdoc__` "protocol" is extended. Please advise.

pdoc/html_helpers.py
```diff
@@ -145,18 +147,20 @@ def lookup(module, refname, link_prefix):
             return None, None
     if isinstance(d, pdoc.doc.Module):
         return d.refname, module_url(module, d, link_prefix)
+    suffix = '()' if isinstance(d, pdoc.doc.Function) else ''
+    name = d.qualname if get_qualname else d.name
```
kernc: Function references are by default labeled with their [full qualified name](https://www.python.org/dev/peps/pep-3155/), e.g. `C.f` or `D.g` in case of methods. Previously, it was the whole refname (`module.submodule.submodule.whole.hierarchy.class.method`). I think this is better, YMMV. Open to discussion.

Before when these helpers were part of the HTML template, it was much easier to override them (and specify `--template-dir`).

kernc:
Happy to add tests once the premises are agreed upon.

Interestingly, current tests all pass when ran locally.

kernc:
Closed in favor of https://github.com/mitmproxy/pdoc/issues/182. Should anyone want to continue, welcome.

Issue #153, opened 2018-06-03
Docstring flavors


cortesi:
I'm opening this ticket to unify discussion of docstring flavors. We have a wide range of requests for these, including:

- Google style [#101]
- reStructuredText [#111]
- Markdown variants [#64]
- Handling of doctests [#63]
- Epytext [#35]
- UML diagrams [#22]
- ASCII Doc [#108]
- Pre-formatted docstrings [#114]
- Formulas [#87]

I'm going to close all these tickets so that we can explore which direction to take in one place. When I first took on pdoc, I thought we could get away with supporting one and only one format. Having read carefully through all of the related tickets, I've now softened that a bit, and I think that there's an argument for being pragmatic. 

My feeling is that we should clearly cater for doctests and pre-formatted docstrings, and choose a more flexible markdown renderer that includes Github-style tables and the like. Equally, I feel we should clearly not include support for reStructuredText, which is way too large and complicated. I'm undecided about things like Google style and EpyText, which are intermediate.

It's possible that we could use a Github-style type marker for this:

    """raw
       
        This is a pre-formatted docstring.
                     Indentation will be preserved.
    """

And:

    """google
    A Google-style docstring.

    Arguments:
         argname (type): Description.
         secondarg (type : default): Description.

    Returns:
        boolean: True on success.
    """

The downside of this is that it introduces a weirdness to docstrings that make them less nice for both humans in code editors and other documentation tools. Another possibility (fronted in some PRs) is to control this unilaterally for an entire module with a command-line flag. This has the issue that the format is not specified in the code (where I feel it should be), and that many modules will be heterogenous and require support for multiple formats. Thoughts?

peterjc:
In epydoc you could set this in the code at file level with '__docformat__', see rejected PEP258 and other documentation:

https://www.python.org/dev/peps/pep-0258/#choice-of-docstring-format

Reusing this convention seems far preferable to inventing another.

kernc:
Could we have `__docformat__` rather a key in `__pdoc__`? The pep was rejected and the idea of namespaces predates it anyway. :slightly_smiling_face:

armant:
Controlling with a command-line flag seems sensible and pragmatic to me. If I rebase and resubmit https://github.com/mitmproxy/pdoc/pull/101, will it be considered?

mhils:
From what I've read so far I think there'd be no strong objections to just start with a universal command-line flag (for example, `--docformat`). 
[...]

PR #180, opened 2018-11-07
Added '--markup' option.

doloopwhile:
Hello.

We are using pdoc for our internal packages. That's very convenient!

However, Some my coworkers are unwilling for pdoc because pdoc assumes all docstrings are written in Markdown and formats their plain-text docstrings wrongly.

Therefore I added `--style` option:

- `--style=markdown`: to interpret as Markdown (current and default behavior).
- `--style=pre`: just to wrap docstring with \<pre\>.

I bet that `--style` is going to boost the users of pdoc.

I think [Google's style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) and [Numpy's style](https://docs.scipy.org/doc/numpy-1.15.0/docs/howto_document.html#numpydoc-docstring-guide) are also expected by not a few people. I will add this two styles if this pull-request is accepted. (However, Currently, I do not have enough time.)

Thanks.


mhils:
refs #153

doloopwhile:
thanks.
I am going to fork.

pdoc/cli.py
```diff
@@ -99,7 +100,22 @@
     default=8080,
     help="The port on which to run the HTTP server.",
 )
+aa(
+    "--style",
```
kernc: I would prefer `--docstring-style`. Less ambiguous.

pdoc/templates/html_index.mako
```diff
@@ -8,14 +8,14 @@ import pdoc.html_helpers as hh
...
-<h1>Python module list</h1>
+  <h1>Python module list</h1>
   <table id="module-list">
   % for root in roots:
     <tr>
       <td><a href="${link_prefix}${root.name}">${root.name}</a></td>
       <td>
       % if len(root.docstring.strip()) > 0:
-        <div class="desc">${root.docstring | hh.mark}</div>
+        <div class="desc}">${root.docstring | hh.mark(markup)}</div>
```
mhils: typo?

PR #187, opened 2019-04-05
Fixed bad module path generation

ysimonson:
`pdoc` is generating incorrect filepaths for modules, causing links to modules to be broken. This seems to fix the issue on the generated docs for a package I'm maintaining; however, I don't know if this is breaking something else in the process.


mhils:
Sorry for not getting back to you here, this is fixed in the latest release! 😃

PR #190, opened 2019-10-10
Use setup() from setuptools

ecederstrand:
This provides support for building wheels etc.


mhils:
Yes, definitely the right thing. Thanks! :smiley:

PR #202, opened 2021-01-19
[requires.io] dependency update on main branch

mhils:


mhils:
D'oh! requires.io is failing here because of pip's new resolver. Let's disable it in the meantime. :)

PR #216, opened 2021-02-03
Align docstrings with locals/attributes

duijf:
This is a small tweak to the CSS which makes the documentation page slightly more readable. 

It's definitely bikeshedding territory. But to me, it's nicer when the docstrings are aligned with the text of the classes/functions instead of with the border.

<details>
<summary>Before screenshot</summary>

</details>

<details>
<summary>After screenshot</summary>

</details>

Thanks for the great project!

### DIFF
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -230,6 +230,7 @@
 
                 /* Page Heading */
                 .pdoc .modulename {
+                    margin-left: 1rem;
                     margin-top: 0;
                     font-weight: bold;
                 }
@@ -289,7 +290,7 @@
 
                 /* Docstrings */
                 .pdoc .docstring {
-                    margin-bottom: 1.5rem;
+                    margin: 0 1rem 1.5rem 1rem;
                 }
 
                 .pdoc > section:first-of-type > .docstring {


mhils:
Thanks for the PR! 😃 We clearly are a bit in bikeshedding territory here, but that is ok. :) I do unfortunately have to admit that I prefer having a clean straight line on the left.
<details>
<summary>Illustration</summary>

</details>


C++/Doxygen (?):

image


image


Rust:

image


We could also do something like this?

mhils:
Besides, this PR has already been super useful to identify that the triangle marker next to "View Source" is currently missing on Firefox. I'll get this fixed for the next release, thanks :)

duijf:
This is the kind of stuff where there are as many preferences as implementations I guess :sweat_smile:

> My proposal would be that I'll switch to your style for a week or so and then decide. Does that work?

Definitely! Thanks for being open to trying; that's more than I had expected in case you didn't like it immediately. No worries if you end up unconvinced though :)

If you do end up liking it: let me know and I'll fix the tests

mhils:
After look at this for a bit longer, I do have to admin that I prefer the current version. Thanks for the nice PR anyways, hope to see you around! 😃 🍰

PR #222, opened 2021-02-14
fix wrong links in index.html

fweisser:
The `replace`-method in JavaScript does only replace the first "." in the modulename. 

This does not work for modules with depth > 2 (e.g. `module.submodule.subsubmodule`, which would generate `href="module/submodule.subsubmodule.html"`.


mhils:
Excellent catch, thanks! 🍰

PR #232, opened 2021-03-14
Don't show non-public decorators above function signatures

### LINKED ISSUE #230: Don't show non-public decorators above function signatures
#### Problem Description
pdoc-generated docs show applied decorators above function definitions by default, including decorators whose names starts with `_`.

#### Proposal
Since the default pdoc behavior is to not document names starting with `_`, I think the application of such decorators should not be shown in the generated documentation for the function, especially since by default documentation for the decorators themselves won't be generated.

zmoon:
Resolves #230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 # Unreleased: pdoc next
 
- - 
+ - Private function decorators (those starting with "\_")
+   are now hidden by default. (@zmoon)
 
 # 2021-03-10: pdoc 6.4.0
 
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -511,7 +511,7 @@
     {%- endif -%}
 {% enddefaultmacro %}
 {% defaultmacro decorators(doc) %}
-    {% for d in doc.decorators %}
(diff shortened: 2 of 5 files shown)


zmoon:
I guess this change introduced a new case of "minor rendering differences"?

mhils:
Thanks for the PR! This looks excellent in general. 🍰 

Tests are failing on Python <= 3.8 because we don't pick up class decorators. Skimming the 3.9 changelog, I think we're hitting https://bugs.python.org/issue35113 here. I don't think this is our business to fix/backport, if someone wants class decorators in their documentation they can upgrade to 3.9. Can you change the snapshot test to use a function decorator (which should be picked up properly), or move it to the misc_py39 file? Either is fine with me. 😃

mhils (approved):
Seems like you have some secret time traveling abilities. 😄 

LGTM.

mhils:
And merged, thanks again! Is a patch release useful for you or do you work with your local copy anyways?

zmoon:
I'm working towards switching some docs that I build and deploy with GH Actions from pdoc3 to pdoc. But I think there may be some other small things I would like to propose before doing that. So it's up to you! Thanks

mhils:
Sure! Just send me a short ping if a release is helpful, the pipeline is automated and takes two minutes. :)

PR #245, opened 2021-04-10
Logo configurability

mrossinek:
I would like to suggest these very simple improvements to the `index` template which will allow greater flexibility on configuring the logo. You can see an example of the outcome here: https://mrossinek.gitlab.io/cobib/

Currently, the situation is still not ideal because if the user were to overwrite `logo.svg` in order to change the main logo, this would also change the pdoc logo in the attribution. I solve this in my use case by overwriting the CSS code in the `style_content` block but if we could embed a jinja block for the main logo path, that would probably the cleanest option. What do you think about this?


mhils (changes requested):
Thanks for the PR! 😃 
This is a good feature request overall, but I have some comments below.

> Currently, the situation is still not ideal because if the user were to overwrite logo.svg in order to change the main logo, this would also change the pdoc logo in the attribution.

It may just make sense to rename logo.svg to pdoc-logo.svg in general. 👍 

> if we could embed a jinja block for the main logo path, that would probably the cleanest option. What do you think about this?

This is a bit tricky because the logo is also a link (which you'd need to configure), so you end up with a hodgepodge of Jinja2 blocks. Maybe we just explicitly include the image in the logo block and adjust the CSS accordingly?
```jinja2
    {% block logo %}
        <a class="logo" href="https://pdoc.dev">
            <img src="data:image/svg+xml,{% filter urlencode %}{% include "logo.svg" %}{% endfilter %}" alt="pdoc logo">
        </a>
    {% endblock %}
```

mrossinek:
Thanks for the review! I have updated the PR accordingly :slightly_smiling_face: 

> It may just make sense to rename logo.svg to pdoc-logo.svg in general. :+1:

Is this something I should include in this PR or would you rather leave that for another time?

mhils:
Either is fine! 😃 

I'll ship a patch release later today. :)

mrossinek:
> Either is fine! 

I think I'll leave it as is then.

> I'll ship a patch release later today. :)

Awesome! :slightly_smiling_face:

mhils (approved):

PR #251, opened 2021-04-22
add overflow-wrap: break-word; to wrap long lines

jstriebel:
Wraps long lines, e.g. in titles, as shown here:
Before:
Screenshot from 2021-04-22 12-48-01

After:
Screenshot from 2021-04-22 12-47-11

Not sure if that's the best css-selector to place it, though.

### DIFF
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -208,6 +208,7 @@
                     /* enforce some styling even if bootstrap reboot is not included */
                     box-sizing: border-box;
                     line-height: 1.5;
+                    overflow-wrap: break-word;
                     /* override background from pygments */
                     background: none;
                 }
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -548,6 +548,7 @@
         {{ decorators(cls) }}
         <span class="def">class</span>
         <span class="name">{{ cls.qualname }}</span>
+        <wbr>
         {{- bases(cls) -}}:
     </div>
 {% enddefaultmacro %}
@@ -648,7 +649,7 @@
             {%- else -%}
                 {{ part }}
             {%- endif -%}
-            {%- if loop.nextitem %}.{% endif -%}
+            {%- if loop.nextitem %}<wbr>.{% endif -%}
         {% endfor %}
     </h1>
 {% enddefaultmacro %}
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -492,7 +492,7 @@
 {% endblock %}
 {% defaultmacro bases(cls) %}
     {%- if cls.bases -%}
-        (
+        <wbr>(
         {%- for base in cls.bases -%}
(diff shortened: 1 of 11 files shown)


mhils:
Thanks! I'm not a big fan of the break in the middle of the module name. Can we use [`<wbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) instead?

jstriebel:
Sounds good! I added the <wbr> tag for the module-names and also for the classes, since those also produced overflows easily, it works nicely. I'd still leave the word-break in there, to prevent any other overflows. <wbr> tags and similar take precedence anyway, so if it's possible a useful breakpoint will be chosen.

A similar topic: Some signature are also wrapped. It would be much more readable, if the line after the break would also be indented, however the indentation seems to happen on the text-level. E.g. see
Screenshot from 2021-04-22 12-25-37

Could this be parsed and html be used for indentation (e.g. via a margin)? Then the indentation would be kept across line breaks. I've no experience with Jinja2, it's not straightforward for me how this would be done.

pdoc/templates/default/module.html.jinja2
```diff
@@ -547,6 +548,7 @@
         {{ decorators(cls) }}
         <span class="def">class</span>
         <span class="name">{{ cls.qualname }}</span>
+        <wbr>
```
mhils, line 551: This introduces an unnecessary space between bases and qualname, please move it to the line above:

image

mhils (commented):
Thanks! Can you point to other examples where overflow is a problem? `overflow-wrap: break-word;` feels like an ugly workaround. 😄 If you don't have specific examples, I'd suggest we don't add it for now but rather deal with things properly when they occur.

> Could this be parsed and html be used for indentation (e.g. via a margin)?

Pretty-printing signatures is surprisingly tricky, which is why we basically do what Python is doing and then add some whitespace:
https://github.com/mitmproxy/pdoc/blob/41215bcce4708aa53c67bad3b2e786a80d387597/pdoc/doc.py#L917-L981

The proper approach to solve this would be to reimplement that logic in the template and emit matching HTML. It does introduce quite a bit of complexity in the template, but I'd generally be happy to see improvements in that direction. We could also ditch the rather hacky PrettySignature, which would be a win. I'd be happy to see a (different) PR for this! 😃

mhils:
Almost forgot: Please also add an entry in `CHANGELOG.md`! 😃

jstriebel:
> The proper approach to solve this would be to reimplement that logic in the template and emit matching HTML. The .signature attribute is a full inspect.Signature subclass and can be accessed as such.

Sounds good :+1: When I have some time in the next weeks to setup the project and work on it more I'm happy to provide a PR.

> If you don't have specific examples, I'd suggest we don't add it for now but rather deal with things properly when they occur.
:+1: Module and class names were the only problems I found for now

I applied your feedback. Thanks for taking the time to review the PR.

mhils (approved):

mhils:
Awesome, thanks again! 😃

mhils:
You'd need to [update the snapshot tests](https://github.com/mitmproxy/pdoc/blob/main/CONTRIBUTING.md#fixing-snapshot-tests), otherwise this LGTM. 😃

Issue #23, opened 2014-12-25
Search Bar


aviralg:
It would be good to have a search bar for fuzzy searching. Maybe you can draw inspiration from Sublime Text's implementation.

BurntSushi:
A search bar would be pretty cool. I don't have Sublime Text and am not familiar with it at all.

I do quite like how `rustdoc` handles searching though: http://burntsushi.net/rustdoc/csv/

aviralg:
I wholeheartedly agree with you. It seems beautiful.
Pdoc is great btw. I am planning to use it in my open source project for documentation.

BurntSushi:
@aviralg Awesome. :-)

mhils:
I'd very much welcome a PR that adds search functionality. I believe the best approach would be if pdoc would output an index file that could be parsed by [Fuse.js](https://fusejs.io/). This would mean we'd have client-side search without external dependencies that can still be hosted statically. Contributions welcome!

mhils:
I have started with a first proof-of-concept at https://github.com/mitmproxy/pdoc/tree/search, 
but don't have the time to continue at the moment. Happy to have someone take over! 😃

Quantalabs:
[Algolia](https://www.algolia.com/) is a good replacement until the search bar is implemented, and I've used it for some of my projects before, and it's absolutely free.

PR #252, opened 2021-04-22
Proposal: Always add "Home / Module Index / Search" button

jstriebel:
It would be great to have a button leading to the top-level index-page in lower-nested pages by default. Not sure what would be the nicest way visually, I just added a simple text button, but also a home or search-magnification icon could be used, maybe even without any text.

Screenshot from 2021-04-22 14-18-28

A simple icon without text would also prevent more line-breaks:
Screenshot from 2021-04-22 14-18-59

I'm sorry if this proposal is too specific, I just thought it might be useful in the default templates.

### DIFF
--- a/pdoc/templates/default/module.html.jinja2
+++ b/pdoc/templates/default/module.html.jinja2
@@ -150,6 +150,20 @@
                     border-color: var(--text);
                 }
 
+                nav.pdoc .index-button {
+                    display: inline-flex;
+                    align-items: center;
+                    color: var(--text);
+                    border-color: var(--muted);
+                    margin-bottom: 1rem;
+                    margin-right: var(--pad);
+                    float: right;
+                }
+
+                nav.pdoc .index-button:hover {
+                    border-color: var(--text);
+                }
+
                 nav.pdoc ul {
                     list-style: none;
                     padding-left: 0;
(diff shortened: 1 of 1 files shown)


jstriebel:
Having a search bar as proposed in #23 would be even nicer and supersede this.

mhils:
Thanks for raising this! How would you feel about a permanent `Module Index` button replacing the "go one layer up" button on submodules? Maybe that's the more useful button in general.

jstriebel:
> Thanks for raising this! How would you feel about a permanent Module Index button replacing the "go one layer up" button on submodules? Maybe that's the more useful button in general.

Good idea, I think that's more useful. However, I'm not sure if I'd expect to go to the index-page with the search or the modules documentation. Maybe buttons for both? E.g. I'd actually expect https://pdoc.dev/docs/pdoc.html to be the "home" page (or Module Index), but https://pdoc.dev/docs/ provides the search and should be accessible from everywhere (until there's a search bar).

mhils:
> Good idea, I think that's more useful. However, I'm not sure if I'd expect to go to the index-page with the search or the modules documentation.

Ah, yes. That's a very good point to make. Thanks!

Given the circumstances I'd lean towards keeping the current setup We will eventually add a search bar that does things proper, until then I'm inclined not to introduce temporary stopgaps. :)

Thank you anyhow for the PR. I hope you don't mind keeping that around in your local config. 😃

PR #258, opened 2021-05-12
Harden Github Actions Config

mhils:


.github/workflows/main.yml
```diff
@@ -30,13 +33,13 @@ jobs:
       matrix:
         include:
           - os: windows-latest
-            py: 3.9.2
+            py: 3.9.5
```
mrossinek, line 36: @mhils May I suggest that you simply use `3.9` instead? This would remove the need to continuously update the configuration whenever a bugfix release of a Python version occurs.

mhils: @mrossinek: Generally yes, but I'm explicitly hard-pinning patch releases here so that our CI doesn't break. pdoc depends on a few Python internals that may cause slightly different snapshot renderings, so it's useful to be explicit here.

PR #267, opened 2021-05-30
Fix blank lines breaking indented code blocks

Crozzers:
Fixed this:
image

Being rendered as this:
image

This doesn't seem like the cleanest fix, however, it only applies to this specific scenario so it shouldn't mess anything up.

It works by replacing blank lines in code blockquotes with line seperator characters (U+2028), which tricks the markdown parser into rendering the blockquote "properly".
I also added a test case for this scenario.

### DIFF
--- a/pdoc/docstrings.py
+++ b/pdoc/docstrings.py
@@ -78,6 +78,21 @@ def _google_section(m: re.Match[str]) -> str:
             contents += "\n"
     else:
         contents = indent(contents, "> ", lambda line: True)
+        if '```' in contents:
+            lines = contents.split('\n')
+            inside_code_block = False
+            for index, line in enumerate(lines):
+                if '```' in line:
+                    inside_code_block = not inside_code_block
+                elif inside_code_block:
+                    if line[2:].startswith(' ') and lines[index - 1] == '> ':
+                        # if the line, minus the blockquote chars, starts with an indent
+                        # and the previous line was blank
+                        # the blockquote will be considered broken up by said blank line.
+                        # add in a character that doesn't register as a space to trick
+                        # the markdown parser into rendering the blockquote as one item
+                        lines[index - 1] = '> \u2028'
+            contents = '\n'.join(lines)
 
     return f"\n###### {name}\n{contents}\n"
 
--- a/test/testdata/flavors_google.html
+++ b/test/testdata/flavors_google.html
@@ -99,6 +99,9 @@
             <li>
                     <a class="function" href="#example_code">example_code</a>
             </li>
+            <li>
+                    <a class="function" href="#indented_code_blank_lines">indented_code_blank_lines</a>
+            </li>
     </ul>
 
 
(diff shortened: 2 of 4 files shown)


mhils:
```python
>>> print(pdoc.docstrings.convert(    """
Test case for blank lines interrupting code blocks when inside an indented block
    
Example:
	```python
	if True:
		print()

		print()

		print()
	```
    """, "google", None))
```

generates this output, which looks correct to me:

```
Test case for blank lines interrupting code blocks when inside an indented block
    

###### Example
> ```python
> if True:
> 	print()
> 
> 	print()
> 
> 	print()
> ```
```

So I'm tempted to believe this is something that should be fixed in markdown2 and not pdoc, or am I missing something? :)

Crozzers:
That would make a lot of sense and it sounds like a much cleaner fix. I'll do that

mhils:
Fantastic, thanks! We have markdown2 vendored, so in principle we do not need to wait for an upstream merge or release. That being said, my experience with them has only been excellent so far. 😃

Crozzers:
The problem seems to lie in `pdoc.markdown2.Markdown._do_code_blocks` in the line
```python
return code_block_re.sub(self._code_block_sub, text)
```
I ran the following function through pdoc:
```python
def b():
    '''
    Sample description

    Example:
        ```python
        if True:
            print(b())

            print(b())

            print(b())
        ```
    '''
    return
```
and when it got to `_do_code_blocks` it had transformed the text into
```html
<div class="codehilite"><pre><span></span><code><span class="k">if</span> <span class="kc">True</span><span class="p">:</span>
    <span class="nb">print</span><span class="p">(</span><span class="n">b</span><span class="p">())</span>

    <span class="nb">print</span><span class="p">(</span><span class="n">b</span><span class="p">())</span>

    <span class="nb">print</span><span class="p">(</span><span class="n">b</span><span class="p">())</span>
</code></pre></div>
```
Which is to be expected. Then, on the line `return code_block_re.sub(self._code_block_sub, text)` the text is changed to
```html
<div class="codehilite"><pre><span></span><code><span class="k">if</span> <span class="kc">True</span><span class="p">:</span>
    <span class="nb">print</span><span class="p">(</span><span class="n">b</span><span class="p">())</span>

<pre><code>md5-[random hash]
</code></pre>

    <span class="nb">print</span><span class="p">(</span><span class="n">b</span><span class="p">())</span>
</code></pre></div>
```
What seems to have gone wrong is the middle `print(b())` line has been matched against their regex and has registered as another code block due to the fact that it is an indented block surrounded by line breaks. It then gets wrapped in `<pre><code>` tags.

The line `return code_block_re.sub(self._code_block_sub, text)` will replace any sections in `text` that match against `code_block_re` (defined in the `_do_code_blocks` function) with the return of `self._code_block_sub` (wraps the input in the `<pre><code>` tags) which results in what we see above.

So what seems to be the solution is tweaking the regex so that it can look ahead and realize that it's already inside a code block so that it doesn't start nesting them

Crozzers:
I'm going to close this PR. I'll keep looking at that regex but I fear it may be beyond me

PR #269, opened 2021-05-31
Add `Doc.source_lines`

mhils:
We don't use this for pdoc itself, but this can be quite useful to link to specific lines on GitHub etc.

@jedie: Is this API matching what you had in mind, or is there anything we can do make it more convenient?

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -152,6 +152,20 @@ class Doc(Generic[T]):
             return None
 
     @cached_property
+    def source_lines(self) -> Optional[tuple[int, int]]:
+        """
+        Return a `(start, end)` line number tuple for this object.
+
+        If no source file can be found, `None` is returned.
+        """
+        try:
+            obj = inspect.unwrap(self.obj)  # type: ignore
+            lines, start = inspect.getsourcelines(obj)
+            return start, start + len(lines) - 1
+        except Exception:
+            return None
+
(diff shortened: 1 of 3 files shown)


test/test_doc.py
```diff
@@ -84,9 +85,10 @@ class RaisingGetAttr(metaclass=_Raise):
         assert c.members
 
 
-def test_builtin_source_file():
+def test_builtin_source():
     m = Module(builtins)
     assert m.source_file is None
+    assert m.source_lines is None
```
jedie, line 91: maybe a test with existing information would be good?

jedie:
Great. That will help...

pdoc/doc.py
```diff
@@ -151,6 +151,20 @@ def source_file(self) -> Optional[Path]:
...
 
+    @cached_property
+    def source_lines(self) -> Optional[tuple[int, int]]:
+        """
+        Return a `(start, end)` line number tuple for this object.
+
+        If no source file can be found, `None` is returned.
+        """
+        try:
+            obj = inspect.unwrap(self.obj)  # type: ignore
```
jedie, line 162: `inspect.unwrap()` is not needed here, because it's called by `inspect.getsourcelines(obj)`

mhils: Good catch, thanks!

PR #278, opened 2021-06-16
 Allow google docstring section headers to contain spaces

Crozzers:
This PR allows google docstring sections to be parsed if they have multiple words in their headers, EG: "Instance Variables" or "Class Attributes".

Currently, custom headers can be added using the `pdoc.docstrings.GOOGLE_LIST_SECTIONS` constant but if the header contains mutliple words then it is ignored.
This is because I forgot to check whether multiple word headers worked when creating #266, I just assumed it would work like single word headers, so this should fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased: pdoc next
 
  - Fix hot-reloading of included Markdown files.
+ - Allow google docstring section headers to contain spaces
 
 # 2021-06-11: pdoc 7.1.1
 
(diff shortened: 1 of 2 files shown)


mhils (approved):
Excellent, thanks! 🍰 

Can you please add a CHANGELOG entry? Otherwise this looks good to merge.

Crozzers:
Cool. Changelog added to. Sorry for the late reply, I made the commit and then forgot to write a response

mhils:
Thanks!

PR #283, opened 2021-07-26
Add show_source flag to allow for hiding of source code in output

Bruntaz:
This adds a commandline flag `--hide-source`, which disables the source of the documented functions from appearing in the output. I think the view source button is a great feature, but for **many** projects it is not acceptable to distribute the source of the function with the documentation because the source is proprietary and under copyright. For example, at my work we use [Cython](https://cython.org/) to compile and ship Python code to help obfuscate the source, but we also need to document the Python APIs that users will program against.

This is mostly just a passthrough flag to change the behaviour of the Jinja template behind the scenes, but I think that hiding the source is a make-or-break feature for many users (including me) and expecting users to edit a Jinja template is far too big of a hurdle to expect people to actually jump over to use a "completely automatic" library.

I've set it to default to showing the source by default.

### DIFF
--- a/pdoc/__main__.py
+++ b/pdoc/__main__.py
@@ -85,6 +85,11 @@ parser.add_argument(
     action="store_true",
     help="Don't open a browser after the web server has started.",
 )
+parser.add_argument(
+    "--hide-source",
+    action="store_true",
+    help="Hide source code of documented functions",
+)
 parser.add_argument("--help", action="help", help="Show this help message and exit.")
 parser.add_argument(
     "--version",
@@ -153,6 +158,7 @@ def cli(args: list[str] = None) -> None:
         edit_url_map=dict(x.split("=", 1) for x in opts.edit_url),
         template_directory=opts.template_directory,
         docformat=opts.docformat,
+        show_source=not opts.hide_source,
     )
 
     if opts.output_directory:
--- a/pdoc/render.py
+++ b/pdoc/render.py
@@ -29,6 +29,7 @@ def configure(
     template_directory: Optional[Path] = None,
     docformat: Optional[Literal["google", "numpy", "restructuredtext"]] = None,
     edit_url_map: Optional[Mapping[str, str]] = None,
+    show_source: bool = True,
 ):
     """
     Configure the rendering output.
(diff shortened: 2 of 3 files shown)


mhils:
Thanks for the PR! The code changes look excellent, but I'm not sure yet if this is a direction we want to go towards.
So far I have deliberately kept the CLI surface as small as possible. It doesn't take much imagination that we could also add a switch to add a logo, add a switch to pass the version number, add a switch to hide specific members, add a switch to embed MathJax, etc. That being said, I think your point that some folks may not like learning a bit of Jinja2 is fair.

Long story short, let me think about this for a few days. We could also add a short section to the docs with a small example on how to override the view source macro (https://github.com/mitmproxy/pdoc/blob/main/examples/custom-template/module.html.jinja2 already has one). Then again, maybe we do want to add a "render options" argument group and put more stuff there.

Bruntaz:
I appreciate the concern over the commandline interface ballooning into something ridiculous but I wonder if there's a middle-ground to be reached here. I've not thought very much about whether this is actually a sensible implementation or even if it is something `pdoc` is designed to support, but it's something to think about at least:

If you use `pdoc` as a module then the `configure` function is there to set up the options as well. This interface could potentially be extended to accept a dictionary of `{"flag": "value"}` (or accept `**kwargs`), which would then update the `env` object with those values:

```python
def configure(*, 
        template_directory=None, docformat=None, edit_url_map=None, show_source=True,  # Values that have commandline options
        **kwargs  # For config options that either the user wants to use in the default templates or add to their own templates
):
    ...
    env.globals["edit_url_map"] = edit_url_map or {}
    env.globals["docformat"] = docformat
    env.globals["show_source"] = show_source
    for key, value in kwargs.items():
        env.globals[key] = value
```

With this approach, the message could be:
* `pdoc` supports a limited set of commandline options for the most important switches
* `pdoc` supports switches for disabling loads of stuff in the output if you call `pdoc` as a library
* `pdoc` supports outputting in literally any format if you know enough about Jinja2 to subclass or write a new template

With this tiered approach, the module can support: users who just want to generate docs from some existing Python they've found lying around; users who wrote a Python tool with docs (and are therefore comfortable with writing pure Python); and advanced users who either know Jinja already or are willing to learn to get a customised output.

mhils:
Thanks again @Bruntaz! This was really useful feedback and made me reconsider the CLI situation. I've just merged #288, which among other things adds a `--show-source/--no-show-source` toggle. 😃

PR #284, opened 2021-07-27
Fix formatting on Google docstrings that have multiple colons

Bruntaz:
This changes the Google docstring parser to match the first colon on the line rather than the last. This change means that this code:

```
Args:
    my_list: A list which defaults to: a, b, c 
```
will be highlighted as "**my_list**: A list which defaults to: a, b, c"  instead of the current "**my_list: A list which defaults to**: a, b, c".

[Google's styleguide](https://google.github.io/styleguide/pyguide.html#doc-function-args) does not explicitly state what should happen when there are multiple colons, but the description of the syntax implies that the first colon is the significant one because `:` is not a valid character in a Python identifier.

### DIFF
--- a/pdoc/docstrings.py
+++ b/pdoc/docstrings.py
@@ -69,8 +69,8 @@ def _google_section(m: re.Match[str]) -> str:
         contents = ""
         for item in items:
             try:
-                # last ":" on the first line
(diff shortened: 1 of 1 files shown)


mhils (approved):

mhils:
This looks excellent, thanks! 🍰

PR #314, opened 2021-11-24
Add {{ super() }} example for custom-template

sitic:
Adds an custom css example to `examples/custom-template`, demonstrating the use of `{{ super() }}`.


mhils (approved):
This is a great addition, thanks! 😃🍰

PR #329, opened 2022-01-06
Add an option to ignore modules with a regex 

### LINKED ISSUE #310: Ignore packages or subpackages
#### Problem Description

My project is a collection of packages that have the following structure

```
package1
    tests_package
    module1
    module2
    ...
package2
    tests_package
    module1
    module2
    ...
```

I would like to ignore `tests_package` from the automatic documentation generation.

#### Proposal

I would like to be able to specify a collection of paths or better yet a search pattern to be ignored by Pydoc.

#### Alternatives

None

#### Additional context

N/A

joubs:
A user can now add modules prefixed by a "!" to exclude them from the list of modules to process. 

Usage:
``` python

pdoc foo !foo.bar foo.bar.baz
```

Fixes #310

### DIFF
--- a/pdoc/extract.py
+++ b/pdoc/extract.py
@@ -48,9 +48,8 @@ def walk_specs(specs: Sequence[Union[Path, str]]) -> dict[str, None]:
     all_modules: dict[str, None] = {}
     for spec in specs:
 
-        str_spec = str(spec)
-        if len(str_spec) > 0 and str_spec[0] == '!':
-            ignore_pattern = re.compile(str_spec[1:])
+        if isinstance(spec, str) and spec.startswith("!"):
+            ignore_pattern = re.compile(spec[1:])
             all_modules = {k: v for k, v in all_modules.items() if not ignore_pattern.match(k)}
             continue
 
--- a/pdoc/extract.py
+++ b/pdoc/extract.py
@@ -46,7 +46,6 @@ def walk_specs(specs: Sequence[Union[Path, str]]) -> dict[str, None]:
     *This function has side-effects:* See `parse_spec`.
     """
     all_modules: dict[str, None] = {}
-
     for spec in specs:
 
         str_spec = str(spec)
@@ -205,8 +204,8 @@ def _all_submodules(modulename: str) -> bool:
 
 
 def walk_packages2(
-        modules: Iterable[pkgutil.ModuleInfo],
-        module_filter: Callable[[str], bool] = _all_submodules
+    modules: Iterable[pkgutil.ModuleInfo],
+    module_filter: Callable[[str], bool] = _all_submodules
 ) -> Iterator[pkgutil.ModuleInfo]:
     """
     For a given list of modules, recursively yield their names and all their submodules' names.
--- a/pdoc/extract.py
+++ b/pdoc/extract.py
@@ -19,13 +19,14 @@ import types
 import warnings
 from contextlib import contextmanager
 from pathlib import Path
+from re import Pattern
 from typing import Callable, Iterable, Iterator, Optional, Sequence, Union
 from unittest.mock import patch
 
 from . import doc_ast, docstrings
 
 
-def walk_specs(specs: Sequence[Union[Path, str]]) -> dict[str, None]:
+def walk_specs(specs: Sequence[Union[Path, str]], ignore_pattern: Optional[Pattern] = None) -> dict[str, None]:
     """
     This function processes a list of module specifications and returns a collection of module names, including all
     submodules, that should be processed by pdoc.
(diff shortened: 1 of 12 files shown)


mhils:
Thanks for the PR! 😃 :cake:
This looks good overall. What happened to the idea from https://github.com/mitmproxy/pdoc/issues/310#issuecomment-954019271 of just doing `pdoc foo !foo.bar`? `walk_specs` could check if the current spec starts with a `!`, and if so, throw out everything in `all_modules` that matches the pattern. One practical benefit would be that you can do `pdoc foo !foo.bar foo.bar.baz`, which is a bit less cumbersome than a similar regex. Is there a reason why this doesn't work, or is there another reason why you prefer to have an additional command line flag?

pdoc/extract.py
```diff
@@ -46,6 +47,13 @@ def walk_specs(specs: Sequence[Union[Path, str]]) -> dict[str, None]:
     """
     all_modules: dict[str, None] = {}
     for spec in specs:
+
+        str_spec = str(spec)
+        if len(str_spec) > 0 and str_spec[0] == '!':
```
mhils, line 52: Nit: Let's do ` if isinstance(spec, str) and spec.startswith("!")`, that's probably a bit more Pythonic. 😃

pdoc/extract.py
```diff
@@ -75,7 +83,7 @@ def walk_specs(specs: Sequence[Union[Path, str]]) -> dict[str, None]:
                 all_modules[m.name] = None
 
     if not all_modules:
-        raise ValueError(f"Module not found: {', '.join(str(x) for x in specs)}.")
+        raise ValueError(f"Module not found: {', '.join(str(x) for x in specs)}, or all modules ignored.")
```
mhils, line 86: How about `f"No modules found matching spec: {', '.join(str(x) for x in specs)}"` ?

The common case here is to mistype a module name. "Module not found: typo, or all modules ignored" is a bit awkward.

mhils (commented):
Ha, thank you for very much for updating the PR so swiftly! If you could take care of adding a small test I'd be happy to take over the remaining documentation part. 😃

joubs:
You commented too fast for me :) 

First of all thanks for the nice feedback, 
I implemented the initial (CLI option) solution intuitively without strong opinion, and I'm happy with the "!" one. 

I was actually confused at first with handling both Path and Strings as input in `walk_specs`, because in the case of a Path if feels a bit awkward to pop the first character. In your comment above, you suggest to exclude the case "!path/to/module", am I right? 

I definitely want to add some unit test, but I had to ask you first: is there a reason for the presence of (missing) `demopackage2` in `test.testdata.demopackage.__init__.py` ? I wanted to write a test based on the tree structure but the "missing module" raised exception is disturbing my tests.

pdoc/extract.py
joubs: Please see my comment below for this part

pdoc/extract.py
joubs: Fixed

mhils:
> I was actually confused at first with handling both Path and Strings as input in `walk_specs`, because in the case of a Path if feels a bit awkward to pop the first character. In your comment above, you suggest to exclude the case "!path/to/module", am I right?

Yes and no. First, you only get Path objects here if pdoc is used as a library. The CLI will always pass in strings as arguments (which may represent file system paths...). So if someone intentionally passes a Path object, then it should stay just that. Second, we match on the final module name only, so `!path/to/module` would not work anyhow because path/to/ is not part of the module name.


> I definitely want to add some unit test, but I had to ask you first: is there a reason for the presence of (missing) `demopackage2` in `test.testdata.demopackage.__init__.py` ? I wanted to write a test based on the tree structure but the "missing module" raised exception is disturbing my tests.

demopackage2 exists at 
https://github.com/mitmproxy/pdoc/blob/main/test/testdata/demopackage2.py
which is technically next to 
https://github.com/mitmproxy/pdoc/tree/main/test/testdata/demopackage.

Confusing, I know. :-)

pdoc/extract.py
joubs: Fixed

mhils (approved):
This looks great now, thanks again! I will take care of documentation later and then get this shipped. 😃 🍰 

Edit: I see there are some test failures, but those should be easy to fix. Don't worry about it. :)

mhils:
I've added a documentation blurp for https://pdoc.dev/docs/pdoc.html, the CLI, and a CHANGELOG entry. Let me know if you feel that anything is missing our should be improved! 😃

joubs:
All looks good to me thanks!
How about these failing tests? I tried  to update the snapshots but there are remaining errors, would you like me to take a deeper look at it?

mhils:
I'll handle those, no worries. :)

Issue #336, opened 2022-01-20
Help with html customization


zaino:
Sorry for abusing the issue tracker, but I don't know where else to get support from after 2 days of trying.

I want to customize the html layout and contents, and I read the documentation about custom jinja files and the -t command line option.
That's all good, but unfortunately I'm very ignorant when it comes to this tech stack, I need a couple of hints to get going.

The main thing that I'd like to accomplish right now is customizing the list of modules in the index page. I created my custom jinja, but I don't understand how to access modules list and hierarchy. I see in the default that `all_modules` is available, but that's only a list of strings. I thought I could access `doc_objects`, but that doesn't seem to be available.

How do I know what underlying python objects are available and how do I access them to create my own population logic?

Thanks in advance for any help!

mhils:
pdoc < 7.1 used to have a feature where, if invoked without arguments, it would render the module index with _all_ installed Python packages. That posed tricky performance issues while not being particularly useful, so it is now removed. One leftover of that feature is that `all_modules` still only is a list of strings (importing all installed modules would have been too costly). Now that we don't have to worry about this particular performance issue anymore, I'd totally support changing `all_modules` to a `dict[module_name, doc.Module]` dict. IIRC someone asked about this a few days ago as well, let me see if I can quickly cook something up. :)

zaino:
Yeah that would be great.
Can you give me a couple of pointers to understand how variables are made accessible in jinja?
For example, how is `all_modules` available and how is it set? I look at all the default templates searching for something like `{% set all_modules...}` but couldn't find anything. I'd like to learn :)

mhils:
It's passed into the template on the Python side:

https://github.com/mitmproxy/pdoc/blob/3d45349c8812bdae38c4ac864df7caf071bd7e28/pdoc/render.py#L77-L100

zaino:
Ok then getting the dictionary you suggested (I see the value is none now) does require a change to render.py.
I just wanted to make sure I wasn't missing already available entry points, basically :)
I'll be eagerly waiting for something to test!

mhils:
I have a prototype at https://github.com/mhils/pdoc/tree/all-modules. I'm still not sure if the changes will make it in as-is, but please feel free to play around with it:
```
pip install https://github.com/mhils/pdoc/archive/all-modules.zip
```

zaino:
That was fast @mhils!
And I'm happy to report it works.
Granted, I just tested:
[...]

PR #337, opened 2022-01-21
Make module objects available in `all_modules`

mhils:
See #336.


pdoc/render_helpers.py
```diff
@@ -229,6 +229,29 @@ def edit_url(
...
+    """
+    Return the name of the (unique) top-level module, or `None`
+    if no such module exists.
+
+    For example, assuming `foo`, `foo.bar`, and `foo.baz` are documented,
+    this function will return `foo`. If `foo` and `bar` are documented,
+    this function will return `None` as there is no unique top-level module.
+    """
+    all_modules: dict[str, pdoc.doc.Module] = context["all_modules"]
+    shortest_name = min(all_modules, key=len, default=None)
```
Fohlen, line 244: such a nice way to implement this!

pdoc/__main__.py
```diff
@@ -188,17 +188,20 @@ def cli(args: list[str] = None) -> None:
...
     else:
-        all_modules = extract.walk_specs(opts.modules)
-        with pdoc.web.DocServer(
-            (opts.host, opts.port),
-            all_modules,
-        ) as httpd:
-            url = f"http://{opts.host}:{opts.port}"
+        try:
+            try:
+                httpd = pdoc.web.DocServer((opts.host, opts.port or 8080), opts.modules)
```
Fohlen, line 193: that should be `port >= 1000` or otherwise linux is going to say no-no

pdoc/web.py
```diff
@@ -97,24 +97,54 @@ def log_request(
...
         super().__init__(addr, DocHandler, **kwargs)  # type: ignore
-        self.all_modules = all_modules
+        module_names = extract.walk_specs(specs)
+        self.all_modules = AllModules(module_names)
 
     @cache
     def render_search_index(self) -> str:
         """Render the search index. For performance reasons this is always cached."""
-        all_mods = {}
+        # Some modules may not be importable, which means that they would raise an RuntimeError
```
Fohlen, line 110: if I understand correctly this is exactly what https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec was designed for

mhils: Not really, things are a bit more complicated here. We use `find_spec` and friends in `pdoc.extract` to get a list of all modules and submodules, but that does not guarantee yet that all those submodules are actually importable. If you module looks like [this](https://github.com/mitmproxy/pdoc/blob/ad32abc7a5e337e8ab424d173638ee516713151b/test/testdata/import_err_simple.py) `find_spec` will happily take it, but you cannot import it later on.

When pdoc is running as web server, we can do two things:

1. Import all modules on startup, as we would when running with `-o`. The first downside is that this can be a bit of a performance hog and pdoc may need multiple seconds to start up. The second downside is that pdoc picks up _all_ submodules, including platform specific ones (think _windows, _linux, ...). So for some projects you will be greeted with quite a few funny error messages, which isn't ideal either.
2. Only import modules on-demand. The upside is that this is faster, the downside is that this may crash at runtime. Also, some special parts (for example the search index) do require everything to be loaded, which is what we account for here.

I'm still not decided on which approach is better long-term, a strong argument in favor of 1) is that it's conceptually simpler. Then again, a potential multi-second startup time is a bit painful as well. I've thought about adding a progress bar to ease the pain, but then again we don't want additional dependencies and we don't want our own progressbar implementation. 😛

pdoc/__main__.py
mhils: Not sure if I'm following here. It's whatever the user specified or 8080 by default. If you explicitly specify something <1024 on Linux then yes, you will get an error unless you have the required extra permissions.

PR #344, opened 2022-02-08
Fix blank lines breaking indented sections within fenced code blocks

Crozzers:
Hello again!

I've decided to take another crack at this problem. This time I think the fix is a bit more promising.
As mentioned previously (in #267), the problem was in the `pdoc.markdown2.Markdown._do_code_blocks` function. The regex expression there would attempt to detect code blocks.

It did have a regex look-ahead to prevent matching against text already inside a code block, which looked like this:
```
(?![^<]*\</code\>)
```
However, this will only check the next HTML tag to see if it is a `</code>` tag, and then stop. Since every line of markdown rendered code usually contains multiple `<span></span>` tags, this does not work. I have altered this look-ahead to:
```
(?![^*]*\</code\>)
```
Which checks the entire rest of the string for `</code>` tags.

### DIFF
--- a/pdoc/markdown2/__init__.py
+++ b/pdoc/markdown2/__init__.py
@@ -1927,9 +1927,11 @@ class Markdown(object):
             ((?=^[ ]{0,%d}\S)|\Z)   # Lookahead for non-space at line-start, or end of doc
             # Lookahead to make sure this block isn't already in a code block.
             # Needed when syntax highlighting is being used.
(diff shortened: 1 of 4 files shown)


mhils:
Thanks! Could you please open a PR at https://github.com/trentm/python-markdown2 to get the changes into the upstream project? I don't like maintaining a custom set of pdoc-specific patches if it can be avoided.

PR #349, opened 2022-02-13
Skip test_formatannotation_still_unfixed test on Python >= 3.10.1

Dliwk:
#### Description
`inspect.formatannotation` was [fixed][1] and it seems to have appeared in Python 3.10.1+. (Currently CI uses maximum 3.10.0).

I'm curious why snapshots have `x: list` annotations instead of `x: list[int]` if they use `formatannotation` from `doc_types.py` which have a fix :thinking:.

[1]: https://github.com/python/cpython/commit/d02ffd1b5c0fd8dec6dd2f7e3f2b0cfae48b7899

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,13 +33,13 @@ jobs:
       matrix:
         include:
           - os: windows-latest
-            py: 3.10.0
+            py: 3.10.2
           - os: macos-latest
-            py: 3.10.0
+            py: 3.10.2
           - os: ubuntu-latest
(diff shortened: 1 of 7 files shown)


Dliwk:
Though it'd be better to just skip that test... and I will be glad if someone explains to me why CI renders exactly `x: list`.

mhils:
Thanks for the PR! :cake: 
The snapshot tests expecting `list` and not `list[int]` is definitely unintended. I'll take a look. 😃

PR #350, opened 2022-02-13
Favicon

denised:
Added the favicon option as discussed.

The most interesting part was realizing that we need a media type to go with it.  What I currently have implemented is a fairly fast and dirty function that will extract the type from the filename, for the most common file types only.  It would fail on other file types, or on URI's that had query parameters (and probably other cases I haven't thought of).  You may prefer something more robust, and/or exposing yet another option to allow the user to set the type themselves. 

I also added a demonstration of overriding the {% favicon %} block to the custom-template example, which shows how to recreate the inlined svg favicon that was originally in use.

Finally, I made one of the snapshot examples use the favicon option (and updated the outputs of all snapshots).


denised:
Dunno what's going on with demo_eager --- I will check on it tomorrow.

mhils:
Thanks for the PR! This was really useful, I would have missed a few things otherwise. 😃 
According to https://html.spec.whatwg.org/multipage/semantics.html#attr-link-type, `type` is advisory, so that's a nice excuse to get rid of the `image_type` complexity. I've also wired up a `--favicon` option and marked the default `favicon.svg` for deprecation if custom templates still reference it. 
Let me know if this looks good to you, I'd be happy to cut a pdoc 10 release once this is in. :)

denised:
Looks good to me --- ship it!  :-)

PR #366, opened 2022-03-14
Module.members: Document intended semantics around private members

jinnovation:
The [documentation](https://pdoc.dev/docs/pdoc.html#control-what-is-documented)
states:

> If __all__ is not defined, then pdoc will consider all members public that do
not start with an underscore and are defined in the current module (i.e. they
are not imported).

However, the `Namespace.members` attribute is intended to be comprehensive, i.e. perform none of the aforementioned filtering. This can be misleading, e.g. in programmatic use of `pdoc`.

As such, this PR clarifies the semantics of the `.members` attribute accordingly.

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -230,7 +230,11 @@ class Namespace(Doc[T], metaclass=ABCMeta):
 
     @cached_property
     def members(self) -> dict[str, Doc]:
-        """A mapping from all members to their documentation objects."""
+        """A mapping from all members to their documentation objects. This includes
+        'private' members; filtering is done elsewhere, e.g. at the templating
+        layer.
+
+        """
         members: dict[str, Doc] = {}
         for name, obj in self._member_objects.items():
             qualname = f"{self.qualname}.{name}".lstrip(".")
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -230,9 +230,9 @@ class Namespace(Doc[T], metaclass=ABCMeta):
 
     @cached_property
     def members(self) -> dict[str, Doc]:
-        """A mapping from all members to their documentation objects. This includes
-        'private' members; filtering is done elsewhere, e.g. at the templating
-        layer.
+        """A mapping from all members to their documentation objects. 
+        
+        This mapping includes private members; they are only filtered out as part of the template logic.
 
         """
         members: dict[str, Doc] = {}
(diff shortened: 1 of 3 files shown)


jinnovation:
Looks like snapshot tests are failing due to the presence of `_PrivateClass` in the output, which -- if we accept that underscored members should not be documented -- is unexpected and undesirable.

Maintainers: would love recommendations on how to proceed here. 👍  Given the changes I made to `pdoc.doc.Module` itself this is a little unexpected to me.

mhils:
Maybe I'm missing something, but if I add `_PrivateClass` to the snapshot and rerun `python test/test_snapshot.py`, the only change to the snapshot is in the highlighted source code. `_PrivateClass` does not show as part of the docs for me.

Maybe there is some confusion because `Module.members` _does_ include private members and they are only filtered out as part of the Jinja2 template [here](https://github.com/mitmproxy/pdoc/blob/4af03e09d760941ed6fdb1ae8505faf0ffa4a9ac/pdoc/templates/default/module.html.jinja2#L228-L251)?

jinnovation:
> Maybe I'm missing something, but if I add `_PrivateClass` to the snapshot and rerun `python test/test_snapshot.py`, the only change to the snapshot is in the highlighted source code. `_PrivateClass` does not show as part of the docs for me.

Ah, that's an oversight on my part then. I didn't notice that it only shows up on the source and that docs don't get generated.

> Maybe there is some confusion because `Module.members` _does_ include private members and they are only filtered out as part of the Jinja2 template [here](https://github.com/mitmproxy/pdoc/blob/4af03e09d760941ed6fdb1ae8505faf0ffa4a9ac/pdoc/templates/default/module.html.jinja2#L228-L251)? Can you clarify what the actual problem is? :)

Ah, good to know. Is this intended behavior, then? That `.members` include **all** members, private or public?

I'm using `pdoc` programmatically, so the inclusion of private stuff in `.members` surprised me. But if this is intended, then I'll happily update my own script to filter out underscored members myself. 

In that case, maybe instead of this PR we could update the documentation for `Namespace.members` to reflect that both private and public stuff is returned? Currently it reads.

> A mapping from all members to their documentation objects.

Either way, feel free to close this particular PR if this is in fact intended behavior.

mhils:
Fantastic, thanks! 🍰

PR #377, opened 2022-04-09
add target to "built-with" link

jacksund:
Adds part 1 of #375 

This simply opens the pdoc website in a new tab when the "built with pdoc" link is selected.

PR #378, opened 2022-04-09
Add target to "built with" link 

mhils:
@jacksund: Updating the snapshots requires Node.js because the snapshots have precompiled search indices (see https://pdoc.dev/docs/pdoc/search.html). Your system didn't have that, which is why you produced different snapshots. We now exit early if node cannot be found. :)

supersedes #377


jacksund:
Gotcha, thanks for fixing this and adding the check condition

PR #379, opened 2022-04-09
shorten submodule text

### LINKED ISSUE #374: Use shortened submodule names by default
#### Problem Description

For packages that have long or verbose import names, the html sidebar can become difficult to read for lower-level modules.

A live example of this is with some documentation in my own package (see [here](https://jacksund.github.io/simmate/simmate/database/base_data_types.html), the "Submodules" list in the sidebar is difficult to read).

#### Proposal
Regardless of the length for full import path, the submodule list can be clear with just a terminal ending. I suggest pdoc uses the submodule name in the sidebar, rather than the full import path. 

As arbitrary example, and sidebar link such as  ...

```
example.path.to.my.submodule1
```

... would be changed to ...

```
submodule1
```

#### Alternatives
But for those that still want the full path visible, an option to incorporate this might be worth considering. Alternatively, the Submodule heading in the html could also list a subheader with the full import path such as...
``` html
<div>

    <h2>Submodules</h3>
    <h4>path.to.parent.module</h4>
    
     <!-- then iterate through submodule list per usual -->
     <p>submodule1</p>
     <p>submodule2</p>

</div>
```

#### Additional context

I know my import paths can be shortened (e.g. changing `base_data_types` to `base`), but I keep them verbose to help with my users -- which have typically never coded or used python before. So this feature change would be super helpful to my project!

jacksund:
fixes #374 

Note, I haven't added functionality to keep the original verbose names. This PR forces the new format of short names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
  - Fix catastrophic backtracking in a markdown2 regex that processes pyshell examples.
    ([#376](https://github.com/mitmproxy/pdoc/issues/376), [@Andrew-Sheridan](https://github.com/Andrew-Sheridan) and [@mhils](https://github.com/mhils))
+-  pdoc now uses the submodule name in the rendered sidebar, rather than the full import path. 
+   ([#374](https://github.com/mitmproxy/pdoc/issues/374), [@jacksund](https://github.com/jacksund))
 
 # 2022-04-06: pdoc 11.0.0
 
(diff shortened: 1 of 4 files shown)


mhils:
This looks good to me. Would you mind adding a CHANGELOG entry? :)

jacksund:
looks like the CI is still going wacky -- even on the main branch. Not sure if I messed things up, but sorry to start this 😓

jacksund:
@mhils I rebased and also updated snapshots with nodejs installed. Any idea where I'm messing things up?

mhils:
Yeah - I accidentally pushed an incomplete fix and broke main myself. 😅 Sorry - should be all set now.

Issue #375, opened 2022-04-08
Improve Rendering of `<h1>` etc. in Docstrings


jacksund:
This issue is just minor suggestions for the default HTML template used by pdoc. These are just preferential suggestions, so feel free to ignore or close this issue if you'd like!

1. Have the "built with pdoc" link open a new tab. When a link switches to a new website it's common practice to use `target="_blank"` to open a new tab rather than update the current one. This would be a super quick fix for pdoc:
``` html
<a href="..." target="_blank">built with pdoc</a>
``` 

2. Add left-side padding under each documentation header. Under a class method, the "property" or "returns" headers are often larger text than the method name itself -- so at a glance, it's difficult organize what belong to what. It'd be much nicer if left-side padding was added to the `<div>` of each class/method/property heading. For example:
``` html

<div>
<h3>MyExampleClass</h3>
    <div style=padding-left: 35px>
          <p>My rendered markdown docstring for this class!</p>
          <h3>def my_example_method(...)</h3>
               <div style=padding-left: 35px>
                     <p>My rendered markdown docstring for this method!</p>
               <div>
    <div>
</div>
```

I'd be happy to implement these as a pull-request if they're accepted!

jacksund:
whoops, accidentally hit submit before I finished writing the issue. Sorry about that. I'm editting the initial comment above to fix this

jacksund:
Updated issue has been added. Sorry about that.

jacksund:
For suggestion 1:
Switch would only involve changing this one line:
https://github.com/mitmproxy/pdoc/blob/995f4f0c02ba71cdb459886346b72eede8d23464/pdoc/templates/default/module.html.jinja2#L64
It would change to:
``` html
<a class="attribution" title="pdoc: Python API documentation generator" href="https://pdoc.dev" target="_blank">
```

mhils:
Absolutely no objections to 1), do you have a screenshot for 2)? I'm not exactly sure what you're referring to.

jacksund:
# Current implementation

![image](https://user-images.githubusercontent.com/47992949/162525393-ed7633b4-1a0c-416d-8aef-e33961509ac1.png)

For the first time a user is ever looking at pdoc-style documentation, it's not immediately obvious where the "Parameters" section belongs to. In fact, it looks like "Parameters" is a new section header for the "def execute" entry. This is because the "Parameters" text size is larger than "def execute" and because both entry are at the same indent level. 

# Suggested implementation

[...]

PR #383, opened 2022-04-20
reduce heading font size within docstrings

mhils:
This PR reduces the font size of headings in docstrings to more reasonable sizes. On top of that, I've been experimenting with different indentations as discussed in #375. Here are the three candidates for now:

| A | B | C |
| - | - | - |
image | image | image

Still haven't made up my mind, feedback welcome. 😃


mhils:
I've polled a few folks at work, the majority prefers B. I think it's a reasonable compromise, let's go with it! 😃

jacksund:
awesome! thanks again for implementing this :partying_face:

martinbenes1996:
Yep, B is the best.

PR #392, opened 2022-05-08
Create .pre-commit-hooks.yaml

rbroderi:
add .pre-commit-hook to allow running with pre-commit.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+-   id: pdoc
+    name: pdoc
+    description: 'pdoc: Auto-generate API documentation for Python projects'
+    entry: pdoc
+    language: python
+    language_version: python3
+    require_serial: true
+    types: [python]


mhils:
Thanks for the PR! I'm a bit reluctant to merge this PR for two reasons:

1. Generally speaking I think it's better to keep rendered docs out of the repository and have a CI job that builds them (see https://pdoc.dev/docs/pdoc.html#deploying-to-github-pages). A pre-commit script encourages setups that are actually unrecommended.
2. I don't use pre-commit myself, and I don't want to take over the maintenance duties for the bindings.

If you would be willing to maintain this hook, I'd be more than happy to point people to a repo of yours if this topic comes up again. I think it just needs a `pyproject.toml` with mitmproxy as a dependency, and the `.pre-commit-hooks.yaml` file. 😃

rbroderi:
Understood. I will take your advice under consideration. If I decide to
create a hook I will let you know. Thanks :-)

mhils:
Thanks for your understanding, and thanks for going through the effort and proposing this! 😃😊

PR #398, opened 2022-05-24
Import mistake

SimonL888:
throwed an error "module not found"

### DIFF
--- a/pdoc/render_helpers.py
+++ b/pdoc/render_helpers.py
@@ -8,8 +8,9 @@ from collections.abc import Collection, Iterable, Mapping
 from contextlib import contextmanager
 from unittest.mock import patch
 
-import pygments.formatters.html
-import pygments.lexers.python
+from pygments import lexers
+from pygments import formatters
+
 from jinja2 import ext, nodes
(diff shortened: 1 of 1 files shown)


mhils:
The existing code works fine for me and on CI. Could you detail where/how you observed an error?

SimonL888:
> The existing code works fine for me and on CI. Could you detail where/how you observed an error?

On using "import pdoc" i get an error "pygments has not attribute lexers" on Pytjon 3.8.9 after I changed the file it worked again.

mhils:
Which pdoc and pygments version do you have installed? You should be able to figure that out by running `pip freeze`.

SimonL888:
Not sure but could be related to this: [](https://stackoverflow.com/questions/12229580/python-importing-a-sub-package-or-sub-module)

mhils:
Could you try reinstalling pygments and/or pdoc? Are you using plain Python, or some Anaconda/... distribution?

mhils:
fixed in 2cd714543c58823c06b74afdddc3c55e24c31a0d

PR #423, opened 2022-08-02
Documentation: fix #422.

### LINKED ISSUE #422: Update documentation of pdoc.doc.Class with info about __init__ on abstract base classes.
#### Problem Description

I apologize if this is not the right template to use. I wasn't sure if I should use the bug report or proposal, but bug report seemed closer.

I am working on a python project that will be packaged as a library, and I am using numpy style docstrings with the constructor documented in the class docstring, so I do not have a docstring on the `__init__` method of my classes, and I am extending the default template to condense the constructor and class declaration into a single signature that reflects usage, but currently, the `pdoc.doc.Class` implementation of `own_members` explicitly removes `__init__` from the returned list when the class is an abstract base class (see #273).

The comment in the source code: https://github.com/mitmproxy/pdoc/blob/d444b885126fbce432bc168fae1cbaa041bda767/pdoc/doc.py#L652 as well as the issue referenced above make it clear that this is intended behavior, and this is easy to work around by adding a docstring to the `__init__` method of the abstract classes that define a constructor as part of their API. However, this workaround is not clearly documented, and the code block that removes the `__init__` method is in the `Class._member_objects` non-public method, which is called indirectly via `Class.own_members` > `Namespace._members_by_origin` > `Namespace.members`, which takes some poking around the source code to identify.

Since this is expected behavior, but it can produce unexpected results in this particular edge case, I think the documentation should be updated to indicate this more prominently.

I think adding this info to the docstring of `pdoc.doc.Class.own_members` would work since that's the portion of the public API that will likely be called when this problem arises.

1. clone the minimal repro: `git clone https://github.com/GrammAcc/pdoc-init-abstractclass-repro.git`
2. cd into the local repo: `cd pdoc-init-abstractclass-repro`
3. install pdoc into the local repo: `pip install pdoc`
4. run pdoc with the custom template and class_signatures module: `pdoc -t . class_signatures.py`

The webserver should open up the documentation page showing two class signatures.
The expected output based on the custom template is for both signatures to show both `arg1` and `arg2`, but the one for the class `NoInitDocstring` shows no arguments as if it did not have an `__init__` method defined. After reviewing the source code and searching for issues on github, its clear that this behavior is intended, but it should be more clearly documented.

#### System Information
pdoc: 12.0.2
Python: 3.10.5
Platform: Linux-5.18.9-arch1-1-x86_64-with-glibc2.35

GrammAcc:
This adds a single docstring to `pdoc.doc.Class.own_members` in order to make the behavior regarding `__init__` methods on abstract classes more clear to users.

I believe the docstring should be concise enough, but I did use a note admonition, which is not used anywhere else in the documentation that I know of, so if the style/formatting needs to be changed, please let me know!

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -629,6 +629,15 @@ class Class(Namespace[type]):
 
     @cached_property
     def own_members(self) -> list[Doc]:
+        """A list of all own (i.e. non-inherited) members.
+
+        .. note::
+            The `__init__` method of an abstract class will not be included in 
+            the list returned by `own_members` if it does not have a docstring. 
+            See https://github.com/mitmproxy/pdoc/issues/273. This can cause 
+            unexpected results in some edge cases. See 
+            https://github.com/mitmproxy/pdoc/issues/422 for more details.
+        """
         members = self._members_by_origin.get((self.modulename, self.qualname), [])
         if self.taken_from != (self.modulename, self.qualname):
             # .taken_from may be != (self.modulename, self.qualname), for example when


mhils:
Thanks! 🍰

PR #463, opened 2022-11-12
Improve conversion of ReST to Markdown by converting the text for :me…

vsajip:
…th:`foo` or :func:`foo` references to `foo()` rather than just `foo`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,14 @@
  - `pdoc.doc.Doc.members` now includes variables without type annotation and docstring.
    They continue to not be documented in the default HTML template.
    ([#107](https://github.com/mitmproxy/pdoc/issues/107), @mhils)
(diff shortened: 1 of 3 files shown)


vsajip:
This is entirely cosmetic, but IMO an improvement as the user can see straight away (without following the link) if the reference is a function or method.

mhils:
Looks good to me. Can you please add a changelog entry and give `replacer` a slightly more meaningful name? Something like `replace_reference`. :)

vsajip:
By the way, after pushing the last commit, I noticed what might be a typo in an older changelog entry:
https://github.com/mitmproxy/pdoc/blob/af7f3d427be77f31d892948537d5855f7427c745/CHANGELOG.md?plain=1#L17
Should that `poc` be `pdoc`?

mhils:
poc -> pdoc is fixed. Thanks again for the PR! 🍰

Issue #466, opened 2022-11-17
Extend Google list sections with more section types


mensonen:
The Google list sections recognises currently only _"Args"_, _"Raises"_ and _"Attributes"_ as headers for sections that consist of a definition list. This could be extended to recognise more section headers, similar to what [sphinx already does](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections).

The list of sections that sphinx supports is quite extensive, but I'd suggest at least adopting "Keyword Args", "Keyword Arguments", "Yields" and "Examples". The "Yields" header is specifically mentioned by Google as an alternative to "Returns", and I have seen "Examples" mentioned in older versions of the style guide. The keyword arguments are not specified by Google as such, but are widely used.

There are also other list-type sections on the Sphinx list, so implementing as much as possible would improve Sphinx compatibility a lot.

mhils:
Contributions are welcome here, the relevant code is in https://github.com/mitmproxy/pdoc/blob/main/pdoc/docstrings.py. :) That being said, my general take here is that type info should live in the type annotation, and then the sections do render fine already. I don't mind additions that come with test coverage though!

ntamas:
I've started addressing an issue similar to this in #489; my use-case is that I have a large codebase spanning several repositories where the docstrings are Google-styled but they use "Parameters" instead of "Args" in the headings. PR #489 could be extended to support the alternatives listed in this issue.

Note that according to my experience, "Yields" is correctly handled by `pdoc` because it is a section whose body is _not_ a list; `pdoc` already converts the header to a Markdown header. "Keyword Args" would need special treatment, and "Keyword Arguments" could be aliases to "Keyword Args" in PR #489.

PR #489, opened 2023-01-03
Add support for alternative section names in Google docstrings

ntamas:
This PR adds support for specifying alternative section names in Google docstrings that are to be interpreted as sections-of-lists, similarly to the standard "Args" section. Right now the PR aliases "Parameters" and "Params" to "Args"; these names are not standard but are often seen in the wild, especially in codebases that mix NumPy and Google-style docstrings (NumPy uses "Parameters" by default).

The PR could also partially fix the issue raised in https://github.com/mitmproxy/pdoc/issues/466 if we decided to add "Keyword Arguments" and "Keyword Args" such that "Keyword Args" is handled as-is and "Keyword Arguments" is aliased to "Keyword Args". That would require an additional commit that I'm happy to do if the maintainer agrees with the general direction of this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@
 
  - Switch from `setup.py` to `pyproject.toml` for pdoc itself. Please file an issue if that causes any problems. 
    ([#474](https://github.com/mitmproxy/pdoc/issues/474), @mhils)
+ - Fix broken links for inherited methods if both parent and subclass have the same name.
+   ([#493](https://github.com/mitmproxy/pdoc/pull/493), @mhils) 
 
  - "Parameters", "Params" and "Arguments" are now also accepted as headings for
    argument lists in Google-style docstrings.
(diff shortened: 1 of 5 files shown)


mhils:
Thanks for the PR! I could see some people being surprised that their sections are renamed by pdoc. Would it make more sense to add those alternatives to `GOOGLE_LIST_SECTIONS` ?

ntamas:
It definitely would; I went for this approach because currently the function also remaps `Args` to `Arguments` so I was thinking that the goal is to use "Arguments" as a sort-of standard section header. If you think that leaving the headings as-is is a better approach, I can update the PR to do so instead.

mhils:
That's actually a fair point. Let's use "Arguments" for everything as you suggested and see if someone complains. While we're at it, would it make sense to add `Arguments` as an alias as well?

mhils (approved):
Awesome, thanks! 🍰

PR #508, opened 2023-02-18
pdoc: fold variable values

### LINKED ISSUE #507: Folding large constants?
First of all, thanks for `pdoc`! I'm a very happy user, both on my hobby projects and professionally.

#### Problem Description

I'm working to port a handful of repositories away from `pdoc3`, for well-trodden reasons. Most of the differences between the two are minor, but one that I'm currently running into is how `pdoc` renders large constants: it renders the entire constant, rather than either truncating or folding it.

For example, this render:

...becomes this in `pdoc`:

(it goes on for quite a bit after that :slightly_smiling_face:)

#### Proposal

It'd be nice to have either a default or a configurable option for truncating or folding large constants! I'm not sure if there's a great heuristic for this, so offering folding (i.e., requiring the user to click a `<details>` element to unfurl) might be slightly better for UX.

#### Alternatives

I assume I could do this with some template hackery on my own, but I figured this would be a reasonable thing to raise upstream!

#### Additional context

No other context. If this is functionality that you'd be interested in, I'd be willing to make an attempt at implementing it. Thanks again!

woodruffw:
Adds a `Variable.source` property similar to the top-level `Doc.source`, ensuring that variables always have a source-like render available.

Also adds the docstring for each variable to the `default_value` render.

Some other notes:

* My editor auto-corrected some whitespace. Let me know if you'd like to avoid those changes, and I can squash them out!
* My solution here (adding `Variable.source`) is arguably a bit of a hack, but it seemed a bit cleaner than the alternative (specializing all of the macros to handle `Variable.default_value_str`)
* I've put the docstring render in the `default_var` macro, but it arguably belongs one level higher. Let me know if you'd prefer that, and I can move it up!

Fixes #507.

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -87,15 +87,15 @@ class Doc(Generic[T]):
     qualname: str
     """
     The qualified identifier name for this object. For example, if we have the following code:
-    
+
     ```python
     class Foo:
         def bar(self):
             pass
     ```
-    
+
     The qualname of `Foo`'s `bar` method is `Foo.bar`. The qualname of the `Foo` class is just `Foo`.
-    
+
     See <https://www.python.org/dev/peps/pep-3155/> for details.
     """
 
@@ -998,7 +998,7 @@ class Variable(Doc[None]):
     default_value: Any | empty  # technically Any includes empty, but this conveys intent.
     """
     The variable's default value.
-    
+
     In some cases, no default value is known. This may either be because a variable is only defined in the constructor,
     or it is only declared with a type annotation without assignment (`foo: int`).
     To distinguish this case from a default value of `None`, `pdoc.doc_types.empty` is used as a placeholder.
@@ -1007,7 +1007,7 @@ class Variable(Doc[None]):
     annotation: type | empty
     """
     The variable's type annotation.
-    
+
     If there is no type annotation, `pdoc.doc_types.empty` is used as a placeholder.
     """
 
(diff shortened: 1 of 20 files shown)


woodruffw:
Before:

image

---

After:



and unfolded:

woodruffw:
I ran `python ./test/test_snapshot.py` locally and updated the snapshots, but I'm still seeing some rendering mismatches:

```
E       AssertionError: Rendered output does not match for snapshot type_stub. Run `python3 ./test/test_snapshot.py` to update snapshots.
E       assert '<!doctype ht...ody>\n</html>' == '<!doctype ht...ody>\n</html>'
E         Skipping 29351 identical leading characters in diff, use -v to show
E         - /span></a><span class="w">    </span><span class="sd">&quot;&quot;&quot;A simple function.&quot;&quot;&quot;</span>
E         ?                ---------------------------
E         + /span></a>    <span class="sd">&quot;&quot;&quot;A simple function.&quot;&quot;&quot;</span>
E         ?           ++++
E           </span><span id="L-8"><a href="#L-8"><span class="linenos"> 8</span></a>
E           </span><span id="L-9"><a href="#L-9"><span class="linenos"> 9</span></a>...
E
E         ...Full output truncated (261 lines hidden), use '-vv' to show

/home/william/devel/pdoc/test/test_snapshot.py:163: AssertionError
```

Any thoughts? :slightly_smiling_face:

mhils:
> Any thoughts? 🙂

It looks like the latest pygments release has slightly altered pygments output. #509 should fix this. :)

(Getting to the rest of your PR later today)

mhils:
Ok - back to the actual PR contents now: Thanks for putting this up so quickly, and thanks for making the changes minimal! This is a neat approach, but I do dislike overloading `.source`. If we could get the actual source code definition here that'd make sense, but I suspect that this is a relatively involved rabbit hole we shouldn't get into right away. I've posted an alternative implementation in #511 that adds a dedicated button. Eventually it'd be nice to have a better display of the long output as well, but for now this fixes the majority of the pain. Let me know what you think, open to additional suggestions!

woodruffw:
Thanks a ton! #511 looks way cleaner than my approach :slightly_smiling_face: 

I'll close this one out.

PR #511, opened 2023-02-19
fold long variable default values, fix #507

### LINKED ISSUE #507: Folding large constants?

mhils:
See #508. 
Screenshot: 
image


woodruffw (approved):
Looks great, thank you!

PR #521, opened 2023-03-15
Added Jinja2 blocks for navigation module, submodule and and index.

### LINKED ISSUE #520: Add {% block nav_index} in module.html.jinja2
#### Problem Description
I would like to do some small modifications to the navigation bar, like renaming the "Contents" header, adding a link to a PDF document etc. As described in the documentation modifications are possible for the logo and the search box, but not for the Contents / Index, Module / API Documentation and Submodule sections. 

#### Proposal
Is it possible (without bad side effects) to enclose the index section of module.html.jinja2 (line 37-41) in a jinja2 {% block}. In that way the user could use the {% extends "default/module.html.jinja2" %} feature as described in the documentation also for the index. 

Blocks for "submodule" (line 43-50) and "module" (line 52-55) would also be quite helpful :)

#### Alternatives
The alternative is for the user to copy the entire module.html.jinja2 file into own repo and modify + use it with explicit -t template. This feels a bit clunky and may introduce problems if there is an update to pdoc where this template is revised substantially.

thanks, Mikkel

mikkelakromann:
fixes #520

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 <!-- ✨ You do not need to add a pull request reference or an author, this will be added automatically by CI. ✨ -->
 
+ - Add additional Jinja2 blocks to allow a more fine-grained customization of the menu.
 
 ## 2023-02-19: pdoc 13.0.0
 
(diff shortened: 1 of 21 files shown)


mhils:
Thanks! 🍰

mikkelakromann:
Well, thank you too!

PR #525, opened 2023-03-26
add: support for in-Markdown Mermaid diagrams

### LINKED ISSUE #524: Add support for additional markdown extras / mermaid
#### Problem Description

GitHub supports rendering Markdown files with Mermaid diagrams, which aren't supported by default with `markdown2` and thus don't get translated correctly.

It seems like a pretty common use case to include your project's README in your module's top-level docstring. Since the diagram doesn't render, I've found that I need to maintain 2 copies of the README, one using the `mermaid` fence block syntax and another with a static `` of the diagram (and making them both an `img` doesn't work since the `img` doesn't respond to GitHub's theme changes).

#### Proposal

This proposal requires bumping the vendored version of the markdown2.

More recent versions have built-in support for rendering mermaid diagrams via the [mermaid extra](https://github.com/trentm/python-markdown2/wiki/mermaid). The extra requires some additional HTML stuff, but I don't think that needs to be included in default template (people can add the necessary JavaScript themselves via a custom template, maybe with a doc explaining how to do it).

#### Alternatives

Rather than always enable the `mermaid` extra and having to maintain the list of `extras` enabled, more recent versions of markdown2 also support enabling extras from within a markdown document via the [use-file-vars extra](https://github.com/trentm/python-markdown2/wiki/use-file-vars). Adding that one extra would let people customize the rendering of their markdowns themselves rather than either fork this project or find another solution.

thearchitector:
Includes support and documentation for rendering Mermaid diagrams by default.

Bumps `markdown2` to `v2.4.8`.

I'd also advocate for listing `markdown2` as a dependency and pinning the version there instead of vendoring a copy. I think it keeps things neater and in their own wheelhouses; I don't see a reason to treat it differently than the other deps, unless there's a proven track record of incompatibilities with other tools people might use in tandem / I acknowledge it would probably add some dependency maintenance overhead (which might be a valid reason itself).

Closes #524.


mhils:
Thank you for the very nice PR! 🍰 I've liked this so much that I moved it into the core and added a `--mermaid` command line flag. 😃 

> I'd also advocate for listing markdown2 as a dependency and pinning the version there instead of vendoring a copy. 

One thing that's special about pdoc is that it gets installed into lots of random virtualenvs with sometimes weird package version constraints. Vendoring makes things easier here. We've repeatedly shipped out-of-band hotfixes for markdown2 (e.g. https://github.com/mitmproxy/pdoc/commit/912c33dc47a6a1e1996039d347da3217ff4f034a), which is why I'd like to retain that capability. 😃

PR #571, opened 2023-06-08
fix possible search bar freezes

Dliwk:
HTML pages, representing modules, include search bar, which replaces all module contents with search results. This cause browser freezes on big modules (example: https://docs.ballistica.net/api7/ba.html, freezes are noticeable when typing first character or when clearing search bar). 

This PR improves this aspect in fact that DOM reflow (which causes freeze) won't happen in the middle of typing the query.

Any ideas, suggestions, etc? Took the 300ms constant out if thin air. Not a JS expert but this seems to be working.

### DIFF
--- a/pdoc/templates/search.html.jinja2
+++ b/pdoc/templates/search.html.jinja2
@@ -8,73 +8,15 @@
     const originalContent = document.querySelector("main.pdoc");
     let currentContent = originalContent;
 
-    function setContent(innerHTML) {
-        {# Replace the entire page contents. Calling this with an empty argument restores the original page. #}
-        let elem;
-        if (innerHTML) {
-            elem = document.createElement("main");
-            elem.classList.add("pdoc");
-            elem.innerHTML = innerHTML;
-        } else {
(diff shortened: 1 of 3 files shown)


Dliwk:
> Could you please not reorder methods? This makes it very hard to review the changes.

It looks like that `let func = ...` statements should be placed before any code that is using it, the only reason I had to put these two methods higher. Am I wrong and same thing could be achieved without reordering?

mhils:
Ah, I see. Ignore my last comment then.

I don't have any significant lag on the page you listed in either Chrome or Firefox. Any hints on how I can reproduce this? If this is not super significant I'd be inclined to keep things fast (without a 300ms delay) for all users.

Dliwk:
I'm currently experience it using pretty low-end hardware, though. (integrated graphics, 2nd gen intel i5). Around 1000ms freeze especially when erasing search bar.

Though even 100ms latency feels different on small pages. Should I close this or maybe you have any thoughts?

Issue #574, opened 2023-06-17
`@property` inconsistently documented


rpitasky:
#### Problem Description
Our codebase uses [a lot of `@property`](https://github.com/TI-Toolkit/tivars_lib_py/blob/fa070d04d0cb8767feeb3273c70fc9776c7c87c1/tivars/var.py#L352-L435) and we're seeing that [only some of these are documented](https://ti-toolkit.github.io/tivars_lib_py/tivars/var.html#TIEntry.data_length). 

Most curious to me is the fact that `TIEntry.data_length` is documented while the other `@property`s are not.

#### Steps to reproduce the behavior:
1. Save the following minimal code to `bug.py`
```
class Bug:
    @property
    def a(self):
        return "a"
```
2. Run pdoc on it (I use `python3 -m pdoc bug.py` but [our CI](https://github.com/TI-Toolkit/tivars_lib_py/blob/fa070d04d0cb8767feeb3273c70fc9776c7c87c1/.github/workflows/docs.yml#L28) does it normally, but I doubt this matters)
3. Observe that the `@property` code is not documented.

#### System Information
Paste the output of "pdoc --version" here.
pdoc 13.1.1, python 3.10

rpitasky:
It looks like the issue is that the @properties don't have docstrings; I find this behavior rather curious.

What is the justification for the default being to not document these?

mhils:
See https://github.com/mitmproxy/pdoc/issues/280 and https://github.com/mitmproxy/pdoc/issues/411 - that should give you some background. I guess I hold no super strong opinions here. :-)

kg583:
I don't have strong opinions about the choice of intended behavior either, but there is still inconsistency.

According to #280, "pdoc documents all attributes that have either a docstring *or* a type annotation." (emphasis mine), but properties are only being documented if there is a docstring *and* type annotation. Indeed, [this property has both](https://github.com/TI-Toolkit/tivars_lib_py/blob/fa070d04d0cb8767feeb3273c70fc9776c7c87c1/tivars/var.py#L353) while [this property is only annotated](https://github.com/TI-Toolkit/tivars_lib_py/blob/fa070d04d0cb8767feeb3273c70fc9776c7c87c1/tivars/var.py#L430), and only former [appears in the docs](https://ti-toolkit.github.io/tivars_lib_py/tivars/var.html#TIEntry.data_length).

mhils:
[...]

PR #575, opened 2023-06-18
Document variables by default

mhils:
Document variables by default, even if they have no docstring or type annotation.

refs #574


rpitasky:
I think configurability here is key, especially considering this is as much of a breaking change as one would see in a project like this.

Issue #210, opened 2021-01-31
Dark theme


Arkelis:
It would be cool if we had a dark theme toggle button, and even better if dark/bright color scheme preference was detected automatically. I might work on this if I have some time.

Congrats for getting the project back on track!

mhils:
Thanks for the kind words! I agree this would be cool, but it also introduces additional complexity I'm not really in favor of.
I'd maybe consider merging a PR that adds *isolated and minimal* dark mode styling based on `prefers-color-scheme`, but that depends on the actual amount of changes required. We definitely won't add an additional button or command line flag for this. 
If a fancier dark mode implementation is super important to you, please feel free to fork and and leave a link to your fork here. It's just that I want to keep pdoc itself simple and lightweight. :)

Arkelis:
I've began to work on this, I can handle color changes with `prefers-color-scheme` media query. The problems come with code highlighting: Pygments does not generate a CSS with color variables, so to have a dark theme scheme, it is needed to have two CSS definitions generated by Pygments. The solutions that I can think of:

- Provide a dark theme by default for code highlighting, so it is suitable for both bright and dark themes. (The simplest)
- Provide a hard coded CSS file for code highlighting just like Bootstrap Reboot instead of creating it with Pygments in the source code: the theme then cannot be easily overwritten without regenerating another CSS files manually with Pygments. But currently we must modify `pdoc`'s source code to change the style of the formatter, using external CSS may be a bit more flexible when using custom templates. (Demands manual usage of Pygments but more flexible)
- Declare two formatters in `pdoc`'s source code and modify the `{% block style_pygments %}` block or the `{{ pygments_css }}` tag. (Simple, but not isolated as you wish)

mhils:
Thanks @Arkelis! 😃 

Option 2 sounds best to me, this is also similar to what [Hugo is doing](https://gohugo.io/content-management/syntax-highlighting/#generate-syntax-highlighter-css) for example. 
Let's add a `syntax-highlighting.css` file and include that instead of `formatter.get_style_defs()`. We can add a short comment in that file explaining how to invoke pygments. I might even be tempted to make the `github` style from https://xyproto.github.io/splash/docs/all.html our light theme default. 😋

mhils:
Fixed in #217!

Aran-Fey:
[...]

PR #594, opened 2023-07-07
Auto dark mode example

nx10:
This PR changes the dark-mode example so that it automatically switches between light and dark mode depending on the users system/browser preference (using the [`prefers-color-scheme` CSS media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)).

It's a very trivial edit and basically just wraps the dark mode theme in
```css
/* default (light) theme CSS */

@media (prefers-color-scheme: dark) {
  /* dark theme CSS */
}
```

chrome_yFx86pezQ9

Related to #210 .

I almost feel like this should be the default pdoc theme, but understand if it is out of scope.


mhils:
Thanks for the PR! My main nuisance with this change is that people will visit https://pdoc.dev/docs/dark-mode/demo.html and it will not be dark. Is there a good way to fix that?

nx10:
Ah I see, yeah I can't think of anything easy to fix that without adding bloat (e.g. adding some JavaScript that e.g. reads an URI parameter, or adding auto-dark-mode as a second example)

nx10:
If you are generally interested in this as a default feature I can also work on a PR that adds a toggle (system default/light/dark) in the top right similar to the [furo sphinx theme](https://pradyunsg.me/furo/quickstart/)

mhils:
Thanks! I do appreciate your effort here. Unfortunately dark mode adds a substantial amount of maintenance overhead, which is something I don't want to take on going forward. As the primary maintainer of pdoc, I don't want the additional burden of testing all changes with multiple color schemes. Thanks again for sending the initial PR, but I'd really like pdoc's scope small here.

nx10:
I see, absolutely understandable. Thanks for your work!

PR #616, opened 2023-09-04
Fix nav horizontal scroll z-index

Domi04151309:
## Description

When embedding content within a page, such as an SVG graphic, with a width exceeding 100%, the vertical scrolling behavior results in an unintended overlay with the menu.

## Fix

Add `z-index: 1` to the `nav` element.

## Before

without

## After

with


mhils:
Thanks! 🍰 

Could you please add a changelog entry? Otherwise this looks good to merge! 😃

mhils (approved):
Thanks! :)

PR #620, opened 2023-09-06
Declare compatibility with Python 3.12

mhils:


mshadbolt:
pydoc isn't compatible with 3.12 though since it depends on distutils which is deprecated in 3.12...

PR #624, opened 2023-09-24
Added parsing of `code-block` ReST directives

JCGoran:
Added simple parsing of `code-block` ReST directives, mainly for compatibility purposes with Sphinx. While in Sphinx there seem to be multiple options for writing down blocks of code, the `code-block` one looks like the one which is, structure-wise, the most similar to Markdown's triple backticks.


mhils (approved):
Thanks! 🍰 😃

Issue #639, opened 2023-11-20
Markdown emphasis captures before math does


andyabateagility:
#### Problem Description
Pdoc appears to keep looking for markdown emphasis tokens `*` within both inline `$...$` and display `$$...$$` math. This make starred environments break Pdoc's math rendering, and leads to unrendered math if matched stars do not have surrounding whitespace.

#### Steps to reproduce the behavior:
For example, this fails to render, and instead produces bare text, missing the stars `*`, and with emphasized text between the stars. Removing the stars allows this to render correctly as displaymath. VSCode's markdown preview renders both correctly.

```
$$
\begin{align*}
f(x) &= x^2\\
     &= x \cdot x
\end{align*}
$$

```

Underscores don't seem to have the same issue, and math renders correctly if there is whitespace around the stars (but that creates other errors in the math renderer, so it can't fix the above issue).

#### System Information
pdoc: 14.0.0
Python: 3.10.12
Platform: Linux-6.2.0-36-generic-x86_64-with-glibc2.35

mhils:
Thanks for the report! The issue here is that we do Markdown processing first, and only then our math renderer (MathJax) kicks in. Eventually we should solve this with a custom markdown2 extra that exempts `$$`/`$` areas from processing. For now, as a workaround, you can escape the `*` to with `\*` get the desired behavior:

```python

r"""
$$
\begin{align\*}
f(x) &= x^2\\
     &= x \cdot x
\end{align\*}
$$
"""

```

andyabateagility:
Thanks for the tip on escaping! Newlines with four backslashes appear to work as well

PR #647, opened 2023-12-12
Docs: clarify that include requires extra escapes

the-rccg:
Including a markdown via `.. include::` results in the markdown requiring \\\ instead of \\ in multiline maths. 
This seems to stem from the escape in Markdown being processed as a string. 
I wouldn't call it a bug, but thought the behavior should probably be documented. Confused me for a while, so here is a hopefully well-understood documentation of it.


mhils:
> requiring \\ instead of \ in multiline maths.

I think you are running into https://github.com/mitmproxy/pdoc/issues/639 here - this does not have much to do with `.. include::`, but with Markdown -> MathJax processing.

PR #654, opened 2023-12-22
show source code of properties

### LINKED ISSUE #653: "View Source" button for @property, @cached_property methods
#### Problem Description
pdoc nicely collects docstrings from both @property as well as @cached_property methods and shows them as documentation for the respective instance attributes. But it doesn't give a "View Source" button to view the source code of the corresponding methods.
I'm not sure if this is intended behaviour. I also couldn't find any directly related issue. Anyway, I'd like to suggest to add a "View Source" button in both cases (and maybe for other related decorators?). I'd say that usually an attribute is a @property (instead of just an instance variable) because there is something interesting inside the @property method, that a reader of the API documentation may well want to see.
I've looked into the source code of pdoc and found the option to produce "View Source" buttons used only in the jija template. But I couldn't figure out:
* do I just need to add similar code somewhere else in the jinja template (where attributes are treated)?
* or is extra handling needed to extract the source code of a @property in the parsing stage of pdoc and associate it to the corresponding attribute? 
Maybe, you could give me a hint, then I'd try and provide a pull request.

Save this to bla.py:

```python
from functools import cached_property


class A:
    """An example class to be documented by pdoc."""

    @property
    def la():
        """Print la."""
        print("la")

    @cached_property
    def li():
        """Print li."""
        print("li")
```

then:

```bash
pdoc bla.py
```

#### System Information

tmeyier:
I think, that pdoc should show the show the source code of properties.

To realize that functionality, the source code of the property function needs to be transferred to the corresponding Variable object, that is used inside the jinja2 template.

The example function bla.py is still in for testing purposes. I'll look into updating the test suite.

closes #653


tmeyier:
I tried to reproduce the lint failures locally. But I get the error below - even with a freshly cloned pdoc. 

I'm using a conda environment with python 3.9, installed the dependencies with conda first and then did
git clone https://github.com/mitmproxy/pdoc.git pdoc
cd pdoc
pip install -e .
tox
Is there something wrong with my setup or is this an issue with pdoc ?

(3.9) thorsten@gustav:~/src/pdoc_original$ tox
lint: commands[0]> ruff .
lint: commands[1]> mypy pdoc test
test/testdata/misc_py312.py:14: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)
lint: exit 2 (0.13 seconds) /home/thorsten/src/pdoc_original> mypy pdoc test pid=18128
lint: FAIL ✖ in 0.17 seconds
[...]

tmeyier:
The line numbers of the @property source don't come out right yet. formatter options need to be set as well.

bla.py
mhils, line 1: This file needs to be removed before this can be merged. :)

pdoc/doc.py
```diff
@@ -282,6 +282,7 @@ def members(self) -> dict[str, Doc]:
                     annotation=doc_f.signature.return_annotation,
                     default_value=empty,
                     taken_from=taken_from,
+                    source=doc_f.source,
```
mhils, line 285: I think this is sensible overall, but as you mention in your comment we should probably also set `source_file` and `source_lines`. _Three_ new attributes in the constructor may be a bit unwieldy, so we could also create the object here and then immediate assign/override those three? Either way - I guess extending the constructor is also reasonable, feel free to make a judgement call here! 😄

mhils (commented):
Looks good! Please also add an entry in CHANGELOG.md. :) 

Normally I'd insist on test coverage, but looks like we already have plenty of examples in the existing tests where this shows, so no action necessary!

mhils:
In terms of workflow: Please feel free to just add more commits here, no rebasing etc. necessary. We'll squash everything into one commit in the end anyhow. :)

bla.py
tmeyier: done.

pdoc/doc.py
tmeyier: I removed the additional attribute from the Variable constructor. Two more seemed too much to me. source, source_file and source_lines are added to the freshly created Variable object now.

mhils (approved):
Awesome, thank you so much for contributing this! 😃 🍰 I thought this would be much more difficult, but you've found a fantastic workaround here.

PR #655, opened 2023-12-23
Introduce @public marker to explicitly mark members as public

### LINKED ISSUE #645: add --dunder flag to include documentation for double-underscore methods

tmeyier:
Here is an implementation of the @public marker, as proposed in #646. 

* members with "@public" in the docstring will be considered public by the default jinja2 template and shown in the generated API documentation - even if the member name starts with _.
* the definition in `__doc__` has higher precedence than @public
* the "@public" marker is removed from the docstring in the rendered docs. This is done with a simple replace filter. So the @public marker should be added directly before the closing """ of the docstring to avoid additional whitespace in the rendered docs.

closes #645

### DIFF
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -462,7 +462,7 @@ You can find an example in [`examples/library-usage`](https://github.com/mitmpro
 from __future__ import annotations
 
 __docformat__ = "markdown"  # explicitly disable rST processing in the examples above.
-__version__ = "14.2.0"  # this is read from setup.py
+__version__ = "14.3.0"  # this is read from setup.py
 
 from pathlib import Path
 from typing import overload
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -184,8 +184,6 @@ ways.
    2. don't have `@private` in their docstring,
    3. and are defined in the current module (i.e. they are not imported).
 
-
-
 In general, we recommend keeping these conventions:
 
 - If you want to document a private member, consider making it public.
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -180,7 +180,7 @@ ways.
 - If `__all__` is defined in the module, then all identifiers in that list will be considered public.
    No other identifiers will be considered public.
 - If `__all__` is not defined, then pdoc will consider all members public that
-   1. do not start with an underscore or have @public in their docstring,
+   1. do not start with an underscore or have `@public` in their docstring,
    2. don't have `@private` in their docstring,
    3. and are defined in the current module (i.e. they are not imported).
 
--- a/pdoc/__init__.py
+++ b/pdoc/__init__.py
@@ -176,18 +176,21 @@ class GoldenRetriever(Dog):
 
 The public interface of a module is determined through one of two
 ways.
-
 - If `__all__` is defined in the module, then all identifiers in that list will be considered public.
    No other identifiers will be considered public.
-- If `__all__` is not defined, then pdoc will consider all members public that
-   1. do not start with an underscore or have `@public` in their docstring,
-   2. don't have `@private` in their docstring,
-   3. and are defined in the current module (i.e. they are not imported).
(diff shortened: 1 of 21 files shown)


pdoc/__init__.py
```diff
@@ -180,9 +180,9 @@ class GoldenRetriever(Dog):
 - If `__all__` is defined in the module, then all identifiers in that list will be considered public.
    No other identifiers will be considered public.
 - If `__all__` is not defined, then pdoc will consider all members public that
-   1. do not start with an underscore,
+   1. do not start with an underscore or have @public in their docstring,
```
mhils, line 183:
```suggestion
   1. do not start with an underscore or have `@public` in their docstring,
```

mhils (commented):
This looks great, thank you so much for tackling this! 🍰

mhils:
Also sorry @public for the notifications! ❤️  Looks like you're taking this with a bit of humor, which is great. Thank you for your understanding, I'll try to go with `@public` going forward! 😅 🍰

public:
I'm used to it :)

test/testdata/visibility.py
```diff
@@ -12,3 +15,36 @@ def my_func():
...
+def my_other_func():
+    """
+    This is another public method that's shown. It should show without additional
+    whitespace above.
+    """
+
+
+def yet_another_func():
+    """
+    This is yet another public method to give a gauge for the expected amount of
```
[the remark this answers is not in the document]
tmeyier: I wanted to see if there is any unwanted whitespace after the docstring  in the generated html. The second method is there only to see how much spacing between methods you get regularly. For the automated test suite it's probably not really useful.

test/testdata/visibility.py
```diff
@@ -12,3 +15,36 @@ def my_func():
...
+    """
+    This is a private method that's not shown because its name starts with __.
+    """
+
+
+def __private_func_explicitly_public():
+    """
+    This is a private method that's shown because it is explicitly marked
+    as public.
+    @public"""
```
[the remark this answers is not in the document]
tmeyier: I don't mind either way. @public at the beginning is also removed without residual white space.

test/testdata/visibility.py
```diff
@@ -8,7 +11,40 @@ class Undocumented:
...
+def public_func():
+    """
+    This is another public method that's shown. It should show without additional
+    whitespace above.
+    """
+
+
+def yet_another_func():
+    """
+    This is yet another public method to give a gauge for the expected amount of
```
tmeyier, line 48:
```suggestion
```

test/testdata/visibility.py
```diff
@@ -1,5 +1,8 @@
 """
-Example where no children should be rendered.
+Example where only the these children should be rendered:
+    * `__private_func_explicitly_public`,
+    * `my_other_func` and
+    * `yet_another_func` should be rendered.
```
tmeyier, line 5:
```suggestion
Example where only the these children should be rendered:
    * `__private_func_explicitly_public` and
    * `public_func` 
```

tmeyier:
The code review functionality in github still is a bit confusing for me. Sorry for the git mess.

mhils:
No worries at all, we'll squash anyways. Thanks again for contributing this, this is really cool! 😃 🍰

PR #657, opened 2023-12-25
get source_file of wrapped functions

### LINKED ISSUE #656: source_file not extracted for cached function
#### Problem Description
I've  created a custom template to show the source file path together with the source of a member.
In this context I found that the source_file of a cached function is not extracted.

```python
# from the root directory of pdoc
from pdoc import doc
from pdoc import extract
from pdoc import render
extract.walk_specs(["test/testdata/demo_long.py"])
d = doc.Module.from_name("demo_long")
d.members["Foo"].members["a_cached_function"].source_file is None
```

#### System Information

pdoc: 14.3.0
Python: 3.12.1
Platform: Linux-6.5.0-5-amd64-x86_64-with-glibc2.37

tmeyier:
I added a function get_source_file to doc_ast.py that tries to get the source file from obj.__wrapped__ if inspect.getsourcefile(obj) doesn't succeed.
I used the same cached approach as in doc_ast.get_source.

closes #656

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## Unreleased: pdoc next
 
+ - pdoc now correctly extracts the source_file of wrapped functions.
 
 ## 2023-12-22: pdoc 14.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Unreleased: pdoc next
 
  - pdoc now correctly extracts the source_file of wrapped functions.
+   ([#657](https://github.com/mitmproxy/pdoc/pull/657), @tmeyier)
 
 ## 2023-12-22: pdoc 14.3.0
 
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -172,10 +172,7 @@ class Doc(Generic[T]):
     @cached_property
     def source_file(self) -> Path | None:
         """The name of the Python source file in which this object was defined. `None` for built-in objects."""
-        try:
-            return Path(inspect.getsourcefile(self.obj) or inspect.getfile(self.obj))  # type: ignore
-        except TypeError:
-            return None
+        return doc_ast.get_source_file(self.obj)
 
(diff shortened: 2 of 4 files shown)


tmeyier:
As an alternative, I thought about unwrapping cached functions in Function.__init__. But I wasn't sure about possible side effects and I found it a bit creepy to test for a cached function with

```python
if isinstance(func, functools._lru_cache_wrapper):
```

tmeyier:
I haven't implemented test coverage for the two new methods in doc_ast yet.

What kind of test would you like to have?
* unit tests inside test/test_doc_ast.py
* an integration test with a custom template using the function source_file (I have one that gives the source_file in the caption of the source code box)?

mhils:
> As an alternative, I thought about unwrapping cached functions in `Function.__init__`. But I wasn't sure about possible side effects and I found it a bit creepy to test for a cached function with
> 
> ```python
> if isinstance(func, functools._lru_cache_wrapper):
> ```

Does that fix source extraction? That seems like a reasonable approach here. We heavily depend on Python internals anyhow. If we're concerned about this being an internal API, we can also try the import in `_compat` and fall back to a placeholder class.

tmeyier:
> > ```python
> > if isinstance(func, functools._lru_cache_wrapper):
> > ```
> 
> Does that fix source extraction? That seems like a reasonable approach here. We heavily depend on Python internals anyhow. If we're concerned about this being an internal API, we can also try the import in `_compat` and fall back to a placeholder class.

I removed the cached version of get_source_file from doc_ast.py and cached functions are unwrapped like this, now:
``` python
        elif hasattr(func, "__wrapped__"):
            unwrapped = func.__wrapped__
```
In the python source, __wrapped__ is declared to be part of the public API of functools, so that should be reasonably safe.

mhils:
Thanks, this looks great now. Can we get a test + changelog entry for this? :)

mhils (approved):
Thanks for the super quick turnaround! 🍰

Issue #442, opened 2022-09-28
Support for typing @overload annotation


jpfeuffer:
#### Problem Description

Some support for function overloads specified by `@overload` annotation in typestubs.

#### Proposal
In the beginning, just one entry per overload would be nice.

#### Additional context
Trying to document a Cython extension that due to its C++ origin has typestubs with `@overload` annotations

mhils:
pdoc heavily relies on dynamic analysis. As far as I know, you cannot access `@overload` signatures using that approach (the final function definition overwrites all previous ones). Implementing support using static analysis would be a major undertaking and introduce significant complexity. I think your request is by all means valid, but I'm afraid the answer here is to 1) use a different documentation generator, 2) not use `@overload`, or 3) live with it.

jpfeuffer:
I see. That makes sense. Do you by chance know about any documentation tool that supports both typestubs and overloads 😅 It is surprisingly hard to find.

Btw: It might be possible starting from 3.11.
https://github.com/python/cpython/issues/89263

mhils:
Seems like this hasn't made it into 3.11 in time and will be in 3.12 the earliest.

> I see. That makes sense. Do you by chance know about any documentation tool that supports both typestubs and overloads

I don't know, but Sphinx and maybe mkdocstrings have the largest ecosystems.

jpfeuffer:
Yes, not in the `help()` function but maybe via the typing stdlib (typing.getOverloads()). The PR for that was merged early enough I think.

Anyway, I give up for now and will revisit the whole thing in the very distant future again. (Sphinx autosummary/doc does not support type stubs at all and overloads only via docstrings and without descriptions of each overload).

Thanks for considering!

PR #665, opened 2024-01-23
Add support for showing all `overload` signatures

daemontus:
This PR should solve #442 by displaying each `overload` type signature as a separate indented block after the "main" method signature.

So far, this works fine on my own project (which uses `PyO3`, so I am also testing it with overloads in stub files). But I am relatively new to `pdoc`, so I am not sure what level of testing is expected for this to be merged. Would it be sufficient to add an overloaded method to `tests/testdata/misc_py311.py` to see how it renders, or is there something more detailed that needs to be done?


daemontus:
Ok, never mind... this still does not work with stub files when the stub file only contains the `override` signatures and the main file contains the "base" declaration... Bear with me, I'll need to test this a bit more...

mhils:
(Happy to look at this once you mark it as ready)

daemontus:
Thanks! Unfortunately, I'll have to come back to this a bit later... the current implementation works as long as both the "base" method and its overloads are all declared in the stub file. But that is not the recommended way of doing things (e.g. `mypy` will report this as an error): the stub file should only contain the overloads, not the base method. 

The issue is that once the overloads are in the stub file without the base method, they don't seem to be accessible through the built-in `get_overloads` function. When debugging, I can see them in the internal data-structures of the `typing` module, but I can't seem to find an "official" way to access them. Maybe it is a bug, but I am not that well versed in Python, so I might just be missing something. Anyway, I'll need to investigate further...

Issue #669, opened 2024-02-21
Link is not working when class is included in `__all__`


nathanthorpe:
#### Problem Description
Links are not working when a class is included in `__all__`.

#### Steps to reproduce the behavior:

Create the following files:

**test_package/client.py**
```python
from test_package.models import Data


class Client:
    def get_data(self) -> Data:
        return Data()
```

**test_package/models/__init__.py**
```python
from test_package.models.data import Data

__all__ = [
    "Data"
]
```

**test_package/models/data.py**
```python
class Data:
    pass
```

Run `pdoc test_package`

Notice the link to the `Data` class is not present.
![image](https://github.com/mitmproxy/pdoc/assets/889094/1efdffc7-3470-47d7-8714-30030ab8a000)

The link works fine if you remove the contents of **test_package/models/__init__.py** and change the import to `from test_package.models.data import Data`.

#### System Information

```
pdoc: 14.4.0
Python: 3.11.3
Platform: Windows-10-10.0.22621-SP0
```

### Possible fix

If `render_helpers.linkify` is changed to what is in [this PR](https://github.com/mitmproxy/pdoc/pull/670), then it displays the link properly (it also removes the `.data` path from the import 

![image](https://github.com/mitmproxy/pdoc/assets/889094/4f9b1e8f-2c0a-4171-9b79-95c544440820)

PR #670, opened 2024-02-21
Fix link to object if it is exposed in `__all__`

nathanthorpe:
Related issue https://github.com/mitmproxy/pdoc/issues/669

Search through the possible sources of the module when looking in `all_modules.`
Also use the path of the matched object.


pdoc/render_helpers.py
```diff
@@ -336,23 +336,27 @@ def linkify_repl(m: re.Match):
...
-            if qualname:
-                assert isinstance(doc, pdoc.doc.Module)
-                doc = doc.get(qualname)
-            target_exists_and_public = (
-                doc is not None and context["is_public"](doc).strip()
-            )
-            if target_exists_and_public:
-                assert doc is not None  # mypy
+            # Check again if the object was re-exposed
+            for module, qualname in possible_sources(
```
mhils, line 340: I _think_ this should use `qualname_candidates` here.

mhils (commented):
Thank you for the PR! I generally agree that #669 is something we should support. 👍 

That being said - we are reaching ugly levels of function complexity here, not sure yet how we can make things simpler. Worst case it just needs more docstrings. I will take a closer look when I have a few cycles. In the meantime, could you please 1) add a snapshot test and 2) a CHANGELOG entry? Thanks!

nathanthorpe:
Thanks for looking into it! 
Ok I believe I've made the appropriate test case. Also added a changelog entry.

pdoc/render_helpers.py
nathanthorpe: `qualname` is a string in this case, so I don't think it should be `candidates`

liborvaneksw:
Hi! I'm facing the same issue. Unfortunately, this PR seems a bit stale. Do you plan to continue with this? Do you need a hand?

nathanthorpe:
This PR should be ready for merge

nathanthorpe:
I just tested it and it looks great, thank you!

PR #671, opened 2024-02-28
Support PEP-0561 compatible stub-only packages

ErikDeSmedt:
Add support for stub-only package.
The typehints for the package `flyingcircus` could be in the `flyingcircus-stubs` package.

### DIFF
--- a/pdoc/doc_pyi.py
+++ b/pdoc/doc_pyi.py
@@ -3,6 +3,7 @@ This module is responsible for patching `pdoc.doc.Doc` objects with type annotat
 in `.pyi` type stub files ([PEP 561](https://peps.python.org/pep-0561/)).
 This makes it possible to add type hints for native modules such as modules written using [PyO3](https://pyo3.rs/).
 """
+
 from __future__ import annotations
 
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


ErikDeSmedt:
I've renamed the `dir` variable to `path_dir`.
There is no impact in this case. However, overriding python built-ins makes me nervous

pdoc/doc_pyi.py
```diff
@@ -25,10 +26,17 @@ def find_stub_file(module_name: str) -> Path | None:
     """Try to find a .pyi file with type stubs for the given module name."""
     module_path = module_name.replace(".", "/")
 
-    for dir in sys.path:
+    # Find .pyi-file in a PEP 0561 compatible stub-package
+    module_part_name = module_name.split(".")
```
mhils, line 30: This won\t work because we replace `.` with `/` a few lines above.

pdoc/doc_pyi.py
```diff
@@ -25,10 +26,17 @@ def find_stub_file(module_name: str) -> Path | None:
...
+    module_part_name[0] = f"{module_part_name[0]}-stubs"
+    module_stub_path = "/".join(module_part_name)
+
+    for path_dir in sys.path:
```
mhils, line 34:
```suggestion
    for search_dir in sys.path:
```
maybe? Strong +1 to not overriding builtins, but `path` in the variable name is typically a `Path` object. I'd be equally happy with just `d`.

mhils (commented):
Thanks for the PR! Could you please add a CHANGELOG entry as well? :)

pdoc/doc_pyi.py
ErikDeSmedt: We don't change `module_name`

```
>>> module_name = "example.submodule"
>>> module_path = module_name.replace(".", "/")
>>> module_name
example.module_name
>>> module_path
example/module_name
```

pdoc/doc_pyi.py
ErikDeSmedt: Fixed

ErikDeSmedt:
Thanks for the quick review. 
I've pushed a new version that should address all comments

pdoc/doc_pyi.py
mhils: Ah, didn't read properly. Thanks! 🍰

PR #684, opened 2024-04-20
Add options for include directive

### LINKED ISSUE #682: Support `start-after` option for `.. include::` directive
#### Problem Description

I want to include my README in the index of my API docs, but this results in a duplicate top-level heading as the README has its own title. It's quite a specific use case but probably one that's likely to recur.

<details><summary>Screenshot</summary>

Screenshot from 2024-04-14 18-17-50

</details>

#### Proposal

Implement the [`start-line` option](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment) for the `include` directive. Then the following code would resolve the problem.

```rst
.. include:: ../README.md
   :start-line: 1
```

#### Alternatives

Users can use CSS to hide the README heading by its `id`, but the navbar still includes the extra heading.

#### Additional context

I'm happy to make/review the PR if you think this is a good idea. For completeness, I can include the other `start-*` and `end-*` options but I think the other options are out of scope.

frankharkins:
Adds support for the `start-*` and `end-*` options of the rst `include` directive. Closes #682.


pdoc/docstrings.py
```diff
@@ -366,11 +366,48 @@ def _rst_admonitions(contents: str, source_file: Path | None) -> str:
     <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html>
     """
 
+    def _rst_extract_options(contents: str) -> tuple[str, dict[str, str]]:
+        """
+        Extract options from the beginning of reStructuredText directives.
+        """
+        matches = re.finditer(r"\s*:(.+?):\s*?(.*?)$", contents, re.MULTILINE)
```
mhils, line 373: Two suggestions here:

1. Options should only appear at the start, not in the middle of the docstring. So this should probable be a `while m := re.match(...):` loop that checks if there currently is an option at the start, processes it, and chops it off.
2. The regex currently isn't great with multiple non-greedy groups next to each other. For example, the second `\s*?` will never match anything. Proposal for an alternative that works with 1):
   ```regex
    ^\s*:(.+?):[ \t]*(.*)([\s\S]*)
   ```
   The last capture group is the remainder of `contents`.

frankharkins: Thanks for the tips! That made the function much neater

mhils:
Thanks! This looks super nice now. Could you please add some basic tests for `_rst_extract_options` and `_rst_include_trim` in [test_docstrings.py](https://github.com/mitmproxy/pdoc/blob/main/test/test_docstrings.py)? Otherwise this looks good to go.

frankharkins:
@mhils done, let me know if it's ok

test/test_docstrings.py
```diff
@@ -26,6 +27,59 @@ def test_rst(s):
     assert not s or ret
 
 
+@given(text())
+def test_rst_extract_options_fuzz(s):
+    assume(not s.startswith(":"))
+    content, options = docstrings._rst_extract_options(s)
+    assert not options
+    assert content == s
```
mhils, line 35: The assertions here are overly restrictive I think. For example, you could have ` :foo:bar` as input, which would correctly parse as an option.

IMO the best approach here is to just skip the assertions, the test then only ensures that we don't unexpectedly raise anything.

test/test_docstrings.py
```diff
@@ -26,6 +27,59 @@ def test_rst(s):
...
+    content, options = docstrings._rst_extract_options(content)
+    assert options == {
+        "alpha": "beta",
+        "charlie": "delta:foxtrot",
+    }
+    assert content == ("\nrest of content\n" ":option ignored: as follows content\n")
+
+
+@given(text())
+def test_rst_include_trim_fuzz(s):
```
mhils, line 54: See above - given that `_rst_include_trim` may also raise for some inputs, we can't really do anything meaningful. I propose we skip this test.

mhils (commented):
This looks great, thanks! Just some minor nits.

mhils (approved):

frankharkins:
@mhils thanks for finishing this up! Sorry I didn't get round to it earlier

mhils:
No worries and thanks again! :)

PR #703, opened 2024-06-25
remove references to polyfill.io

adhintz:
The polyfill.io website has been reported to serve malicious code. ([reference](https://sansec.io/research/polyfill-supply-chain-attack) and [GitHub issue](https://github.com/polyfillpolyfill/polyfill-service/issues/2873#issuecomment-2143228377))

I do not think there is a need for these es6 polyfills because es6 has been supported in browsers for the past 9 years.


mhils:
Thank you for flagging this, greatly appreciated!

adhintz:
Thank you for the quick merge, appreciate it!

mhils:
pdoc 14.5.1 is out. Advisory is at https://github.com/mitmproxy/pdoc/security/advisories/GHSA-5vgj-ggm4-fg62, I've requested a CVE from GitHub. Thank you again for the report! 

I will now do some digging how this made it in the in the first place. I typically avoid CDNs where possible, but I vaguely recall that this was tricky with MathJax.

mhils:
After doing some more digging, we included polyfill.io here because that is what's recommended on https://www.mathjax.org/#gettingstarted. I've flagged the latest developments at https://github.com/mathjax/MathJax-docs/issues/334. 

We're still using cdnjs for MathJax and Mermaid diagrams if those features are enabled (they are off by default). I'm much less worried about cdnjs, but we should take a look again if there is a good way for us to embed both of them into pdoc without massively bloating our output.

mhils:
@adhintz: I've credited you for reporting this over at https://github.com/mitmproxy/pdoc/security/advisories/GHSA-5vgj-ggm4-fg62. There's probably a button for you somewhere to accept that. Thank you again! 🍰

PR #716, opened 2024-07-12
docs: make it more clear what `@private` does

mhils:
/cc @elliot-100 - does this look reasonable?

Also, thanks! :)


pdoc/__init__.py
```diff
@@ -199,6 +199,8 @@ class GoldenRetriever(Dog):
 You can find an example at
 [`examples/custom-template/module.html.jinja2`](https://github.com/mitmproxy/pdoc/blob/main/examples/custom-template/module.html.jinja2).
 
+Hiding an item only removes it from documentation. It is still displayed in the source code when clicking the "View Source" button.
```
elliot-100, line 202: Happy with the wording; I'd suggest moving it up one paragraph though, before 'As a last resort..' but not critical. Thanks.

PR #730, opened 2024-08-13
Add support for Python 3.13

mhils:


mhils:
The only remaining issue here is https://github.com/python/cpython/issues/122981.

Issue #731, opened 2024-08-24
Import resolution for __init__.pyi stub picks wrong package


apparebit:
#### Problem Description

When a typing stub for a module (say, the extension module `prettypretty.color`) uses an `__init__.pyi` file (say, `prettypretty/color/__init__.pyi`) and contains a relative import for a submodule (say, `from . import gamut`), pdoc uses the outer module's package (`prettypretty`) as the package for the submodule instead of the module (`prettypretty.color`) and then tries to load a non-existent module (`prettypretty.gamut` instead of `prettypretty.color.gamut`). More generally, Python's import machinery evaluates `directory/__init__.py(i)` as such, even if it becomes accessible through `sys.modules` as the `directory` module later on.

The corresponding exception for the above example:

```
(.venv) rgrimm@Precious prettypretty % python -m pdoc -o target/doc/pdoc prettypretty.color
Warn: Error parsing type stubs for prettypretty.color:
Traceback (most recent call last):
  File "/Users/rgrimm/Work/prettypretty/.venv/lib/python3.12/site-packages/pdoc/doc_pyi.py", line 118, in include_typeinfo_from_stub_files
    imported_stub = _import_stub_file(module.modulename, stub_file)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rgrimm/Work/prettypretty/.venv/lib/python3.12/site-packages/pdoc/doc_pyi.py", line 52, in _import_stub_file
    eval(code, m.__dict__, m.__dict__)
  File "/Users/rgrimm/Work/prettypretty/prettypretty/color/__init__.pyi", line 5, in <module>
    from . import gamut as gamut
ImportError: cannot import name 'gamut' from 'prettypretty' (/Users/rgrimm/Work/prettypretty/prettypretty/__init__.py)
 (/Users/rgrimm/Work/prettypretty/.venv/lib/python3.12/site-packages/pdoc/doc_pyi.py:120)
```

I checked that what I describe as expected behavior for `from . import spam` is the actual behavior for Python. In addition to directly running pdoc, I also ran pdoc from a small Python script that loads the native library first to exclude any other causes (importing `prettypretty.color` also imports all submodules such as `prettypretty.color.gamut` and makes them available through `sys.modules`). The custom script resulted in the same error.

#### Steps to reproduce the behavior:
1. Clone [prettypretty](https://github.com/apparebit/prettypretty)'s repository
[...]

PR #732, opened 2024-08-26
Improve pyi file import

mhils:
refs https://github.com/mitmproxy/pdoc/issues/731

@apparebit: Could you please check if this fixes your issue? I've tried following your repro steps, but this fails when running `./rr.sh` the second time (maturin: not found).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix a bug where entire modules would be excluded by `--no-include-undocumented`.
   To exclude modules, see https://pdoc.dev/docs/pdoc.html#exclude-submodules-from-being-documented.
   ([#728](https://github.com/mitmproxy/pdoc/pull/728), @mhils)
+- Fix a bug where pdoc would crash when importing pyi files.
 - Fix a bug where subclasses of TypedDict subclasses would not render correctly.
   ([#729](https://github.com/mitmproxy/pdoc/pull/729), @mhils)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   To exclude modules, see https://pdoc.dev/docs/pdoc.html#exclude-submodules-from-being-documented.
   ([#728](https://github.com/mitmproxy/pdoc/pull/728), @mhils)
 - Fix a bug where pdoc would crash when importing pyi files.
+  ([#732](https://github.com/mitmproxy/pdoc/pull/732), @mhils)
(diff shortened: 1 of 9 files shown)


apparebit:
I installed the wheel generated by CI and it does fix the bug I reported.

However, it also surfaces another bug, at least from what I can tell. The typing stubs for pretttypretty contain a good number of self- and forward references. Pylance has no issue with them and they all type check. Mypy's stubgen tool generates stubs like that as well. Hence, I do believe that's acceptable. Alas [PEP 484](https://peps.python.org/pep-0484/#stub-files) uses lots of words to say very little beyond stub files having the "same syntax as regular Python modules." Since pdoc seems to be evaluating the stubs like a regular module loader would, self- and forward references cause errors because no binding can be found. Why does pdoc execute stubs? As a convenience for inspecting the declarations?

apparebit:
`./rr.sh install` uses Rustup to install the Rust tool chain. In turn, Rustup modifies most shell initialization files such as `.profile` and `.bashrc` to update the PATH like so:
```sh
export PATH="$HOME/.cargo/bin:$PATH"
```
The install step should have printed a warning to restart your shell to pick up the changes to the PATH. Can you please check whether the shell initialization files were updated? I want to make sure the runner script is as robust as possible.

mhils:
> I installed the wheel generated by CI and it does fix the bug I reported.

Thanks for verifying!

> Why does pdoc execute stubs? As a convenience for inspecting the declarations?

pdoc is fundamentally based on dynamic analysis, and that extends to pyi files as well. I don't want to maintain a full-blown static analyzer for Python-like files, so dynamic execution is the best we can do. On the upside, you can typically make forward references work with
```python
from __future__ import annotations
```
at the top of your Python file.

> The install step should have printed a warning to restart your shell to pick up the changes to the PATH.

It did, but I didn't pay attention and blindly followed your steps. Thanks for clarifying, seems to work otherwise!

apparebit:
Right, but even an execution-based dynamic analysis can support self- and forward-references rather easily:

1. Extract the names and possibly kinds of all module-level bindings, i.e., the `def`, `class`, and variable definitions.
2. Create a binding to a dummy value for each of the names in the dictionary serving as module environment.
3. `exec` as before.

The dummy bindings ensure that lookups during evaluation don't fail. But by the end of module evaluation all dummy bindings have been replaced with real ones. I even seem to remember that some Python standard library module implements the name extraction already. I'll have a look tomorrow.

mhils:
Happy to take a look at patches. :)

PR #734, opened 2024-08-30
Supply an empty name to `ast.ClassDef` and `ast.FunctionDef`

ld-cd:
Python 3.13 appears to be deprecating not providing the name parameter for `ast.ClassDef`, this patch provides an empty string as a name when an empty `ast.ClassDef` or `ast.FunctionDef` is created.

See:
https://github.com/MazinLab/binney/actions/runs/10627764750/job/29462165838#step:5:310

Signed-off-by: Aled Cuda <aled@ucsb.edu>


ld-cd:
Missed the open PR for this

PR #754, opened 2024-10-23
fix(#753): API documentation for `unittest.mock`

### LINKED ISSUE #753: TypeError: argument of type '_SentinelObject' is not iterable (`unittest.mock`)
#### Problem Description

`pdoc` fails to generate API documentation for the `unittest.mock` module.

```sh
pdoc --output-directory /tmp/docs unittest.mock
```
```py
Traceback (most recent call last):
  File "/private/tmp/pdoc/.venv/bin/pdoc", line 8, in <module>
    ...
  File "/Users/bf/.pyenv/versions/3.13.0/lib/python3.13/functools.py", line 1037, in __get__
    val = self.func(instance)
  File "/private/tmp/pdoc/.venv/lib/python3.13/site-packages/pdoc/doc.py", line 259, in members
    taken_from = self._taken_from(name, obj)
  File "/private/tmp/pdoc/.venv/lib/python3.13/site-packages/pdoc/doc.py", line 448, in _taken_from
    if mod and qual and "<locals>" not in qual:
                        ^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type '_SentinelObject' is not iterable
```

#### System Information

```sh
❯ pdoc --version
pdoc: 15.0.0
Python: 3.13.0
Platform: macOS-14.6.1-arm64-arm-64bit-Mach-O
```

bryant-finney:
# Closes #753 

Resolve the `TypeError` by testing for the `__contains__` attribute

### DIFF
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -445,7 +445,7 @@ class Module(Namespace[types.ModuleType]):
 
         mod = _safe_getattr(obj, "__module__", None)
         qual = _safe_getattr(obj, "__qualname__", None)
-        if mod and hasattr(qual, "__contains__") and "<locals>" not in qual:
+        if mod and isinstance(qual, str) and "<locals>" not in qual:
             return mod, qual
         else:
             # This might be wrong, but it's the best guess we have.
--- a/pdoc/doc.py
+++ b/pdoc/doc.py
@@ -445,7 +445,7 @@ class Module(Namespace[types.ModuleType]):
 
         mod = _safe_getattr(obj, "__module__", None)
         qual = _safe_getattr(obj, "__qualname__", None)
-        if mod and qual and "<locals>" not in qual:
+        if mod and hasattr(qual, "__contains__") and "<locals>" not in qual:
             return mod, qual
         else:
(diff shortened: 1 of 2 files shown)


pdoc/doc.py
```diff
@@ -445,7 +445,7 @@ def _taken_from(self, member_name: str, obj: Any) -> tuple[str, str]:
 
         mod = _safe_getattr(obj, "__module__", None)
         qual = _safe_getattr(obj, "__qualname__", None)
-        if mod and qual and "<locals>" not in qual:
+        if mod and hasattr(qual, "__contains__") and "<locals>" not in qual:
```
mhils, line 448: Do you think
```suggestion
        if mod and isinstance(qual, str) and "<locals>" not in qual:
```
would maybe be more appropriate here?

mhils (commented):
Thanks! 🍰

pdoc/doc.py
bryant-finney: Ah yes, nice catch 👍 (should have checked the signature)

mhils (approved):
Thanks!

PR #763, opened 2024-11-29
Update CDN location of mermaid from v10 to v11

CodeMelted:
I discovered your product and it is awesome. However noticed that mermaid (one of the awesome features you provide) is on `v10`. This will bring it up to `v11`. In the tests I ran for rendering mermaid documents it supports the upgraded mermaid syntax / styling.

### DIFF
--- a/pdoc/templates/mermaid.html.jinja2
+++ b/pdoc/templates/mermaid.html.jinja2
@@ -6,7 +6,7 @@
     }
 </style>
 <script type="module" defer>
-    import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs";
+    import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs";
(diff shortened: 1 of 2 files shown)


mhils (approved):
Thanks! 🍰

joshuai96:
Happy to see the update. Currently seeing `Syntax error in text` with the used mermaid version 10.9.3. But working on mermaid 11.

Maybe adding a parameter to `pdoc` to provide a link to mermaid yourself is a good idea? Like overwriting the CDN link.

- choose your own version (up and downgrade yourself)
- use a self hosted version
  - if you don't want to use a CDN
  - if you don't have access to global internet

mhils:
I've pushed a new release. 

> Maybe adding a parameter to pdoc to provide a link to mermaid yourself is a good idea? Like overwriting the CDN link.

This adds additional CLI complexity, which we want to avoid. If you need a custom Mermaid version, you can always put a `mermaid.html.jinja2` in your custom template directory.

PR #771, opened 2025-01-31
Fix Markup-returning functions' return type

iFreilicht:
Markup is a sub-class of str, so this wasn't strictly *wrong*, but it's more correct this way and clearer when extending these filters.


mhils (approved):
Thanks! 🍰

PR #779, opened 2025-02-17
fix tiny mistake in warning message

a-detiste:
The actual import is correct at the top of this one same file.


mhils (approved):
Thanks! 🍰

Issue #758, opened 2024-11-18
pdoc should display which dostring line is causing the issue if a ValueError is raised on re.split


dhermangh:
#### Problem Description
In Returns section, if the numpy docstring is missing a type description, a ValueError is raised. It is fine, however a message displaing which line raised the error should be displayed, helping catch which docstring lines are incorrect. Sometimes linters miss such errors.
For example such docstring:

```Python
def foobar(x: str) -> Y:
    """
    I am a docstring in a numpy format

    Parameters
    ----------
    x: str
        I'm a parameter of a str type.

    Returns
    -------
        Tuple containing status message string and CPE, time delta or None
    """
```

Will raise said ValueError:
```
Traceback (most recent call last):
  File "/home/qabot/venvs/qabot_venv/bin/pdoc", line 8, in <module>
    sys.exit(cli())
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/__main__.py", line 199, in cli
    pdoc.pdoc(
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/__init__.py", line 529, in pdoc
    out = render.html_module(module, all_modules)
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/render.py", line 106, in html_module
    return env.get_template("module.html.jinja2").render(
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/templates/default/module.html.jinja2", line 311, in top-level template code
    {%- if loop.nextitem -%}
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/templates/default/frame.html.jinja2", line 36, in top-level template code
    {% block body %}
  File "/home/qabot/venvs/qabot_venv/lib/python3.10/site-packages/pdoc/templates/default/frame.html.jinja2", line 42, in block 'body'
    {% block content %}{% endblock %}
[...]

PR #784, opened 2025-03-14
add descriptive error message if re.split fails in numpy

tobiscode:
Hi,
I ran into the same issue as reported in https://github.com/mitmproxy/pdoc/issues/758, and have created a fork to implement the solution proposed by @dhermangh. In its current form, it prints the affected content and the header it's under, but since the individual docstring conversion function does not know which function/class etc. it is operating on, it can't give that info (or a line number in the original file, for that matter). If there is a way to know the latter, I'm happy to add it as well, but for now, it at least suggest to the user that a type descriptor might be missing.
Does this look good to everyone? Happy to include more fixes.
Cheers,
Tobias


mhils:
Thanks! I'd suggest we generically catch all exceptions in `convert`, and then do something like this:

```python3
except AnyException:
    raise RuntimeError(f"Docstring processing failed:\n{content=}\n{source_file=}\n{docformat=}") from e
```
(see pdoc.extract for AnyException - we can move it somewhere else)

mhils:
Instead of. The point of that is to help debugging and fixing pdoc. 

I'm happy to merge a second PR that fixes the actual bug in the numpy docstring conversion.

tobiscode:
>  that fixes the actual bug in the numpy docstring conversion.

I'm not sure it is a bug - the [NumPy docstring convention](https://numpydoc.readthedocs.io/en/stable/format.html#parameters) requires there to be a type for both in- and output parameters. Sphinx has an [extension](https://github.com/tox-dev/sphinx-autodoc-typehints) that takes advantage of python type hints but I think it's just for convenience rather than a newer docstring convention. How would you have suggested to fix the parsing bug otherwise?

tobiscode:
> `pdoc.docstrings.convert` is not supposed to raise an exception under any circumstances. It may produce a warning, but it must not error.

Now I'm even more confused, since in your first comment you suggested to catch `AnyException` in `pdoc.docstrings.convert` and then turn it into a `RuntimeError`. Don't get me wrong, I'm happy to implement whatever solution you prefer, it's just unclear to me what that is.

If you meant `pdoc.docstrings.**numpy**` is not supposed to raise an exception, then I could modify the `re.split` operation such that it throws a warning instead. In that case, I would need to know what the code is supposed to do for that section - just ignore it (in which case the Returns section would be empty) or add a dummy variable name (maybe `_` in keeping with some Python standards, this would allow the description to still appear in the generated documentation).

tobiscode:
Okay now I think I get it, thanks for the clarification. What I'll do then is:

1. move the try/except block from the `numpy` subfunction to the `convert` function, and update this PR with that commit, and
2. make a minimal test case that for the NumPy-specific case mentioned in https://github.com/mitmproxy/pdoc/issues/758 produces different test outputs. I'll write that to the issue first, and then after deciding how to handle it, I'll make a new PR.

mhils:
Perfect, thanks!

tobiscode:
Okay so it seems the test coverage action fails on these lines:

```python
    except AnyException as e:
        raise RuntimeError(
```

How should I address this? By making a test case in `test_docstrings.py` which will raise that exception?

mhils:
Yes please. The test needs to mock one of the subfunctions to raise. pytest's monkeypatch is great for that.

tobiscode:
Okay, so since I wasn't familiar with monkeypatch I imitated how the `AnyException` catches were tested in `test_extract.py` and did something similar in `test_docstrings.py`. I hope that's okay, otherwise I can try to do it the monkeypatch way. With that, all tests pass, and the branch could be merged.

There is still a test warning appearing in the logs, however, which seems unrelated to my changes:

```
 test/test_doc_types.py::test_eval_fail2
  D:\a\pdoc\pdoc\pdoc\doc_types.py:148: UserWarning: Error parsing type annotation xyz for a. Import of xyz failed: name 'xyz' is not defined
    warnings.warn(
```

It looks to me in `doc_types.py` that this is one of the possible outcomes, but not the one it's testing for. I don't know how to fix this since I don't understand the logic flow in this test, but happy to make some changes if someone explains it to me. (And probably make a new PR for that.)

test/test_docstrings.py
```diff
@@ -37,6 +37,15 @@ def test_rst_extract_options_fuzz(s):
     assert not s or content or options
 
 
+def test_convert_exception():
```
mhils, line 40: This is not a useful test because it tests unwanted behavior (we don't want any known cases of raising). Let's use pytest's monkeypatch to replace `rst` with a method that raises an exception, and then test with that.

tobiscode: Ahhh I see, that makes more sense.

tobiscode: Does the new commit look good now?

mhils (approved):
Thanks! 🍰 😃

Issue #282, opened 2021-07-15
Support including images (through reST directives)


StellarStorm:
#### Problem Description

First of all, it's great to see this package being developed again! I've only just begun converting my docstrings into external documentation and this looks like it will be a great fit for what I need!

"A picture is worth a thousand words" and this is sometimes especially true for documentation! Well-written documentation should always be the standard of course, and I'm very glad this project makes it easier to read docs in a browser rather than having to always read the code docstrings themselves. Sometimes, though, being able to include an image for a difficult operation or concept would make a world of difference.

For instance, my research focuses on a subfield of deep learning. Having a way to display small images/graphs of before and after a function is run on an array (or something similar) would go a long way into making certain difficult operations easier to conceptualize for myself and others.

#### Proposal

It would be great if this project added a way to include images into the output HTML files. One way to do this might be to use the reST directive `.. image:: ` which could easily be added to the bottom of a docstring. That way, the docstring itself is minimally disrupted, but when viewing the generated external documentation, images could be easily seen.

Additionally, an extra bonus would be if this supported the google doc style instead of only restructured text. I mention this because [including markdown files](https://pdoc.dev/docs/pdoc.html#include-markdown-files) mentions using the `__docformat__ = "restructuredtext"`, and I assume if adding images were to be supported, it might be done in a similar way. Unfortunately our code follows the Google doc style and setting `__docformat__ = "restructuredtext"` messes up the rendering.

#### Alternatives
I've considered using small Markdown snippets like
```markdown
![image](/path/to/image.png)
```

and then including them in a docstring via
```python
.. include:: snippet.md
```

However, this only displays the text content of the markdown file and not the included image.

#### Additional context

This is something that [pdoc3 says they support](https://pdoc3.github.io/pdoc/doc/pdoc/#supported-rest-directives) through the reST `.. image::` directive so hopefully it wouldn't be too much trouble to also integrate it here. I haven't tried using this feature from that package, but their documentation does say it is an option.

This looks similar to #272, however, I'm only concerned with being able to include images in the static HTML files, not display any warnings or caution symbols. Additionally, I believe (or hope! 😁 ) that this is within the scope of the project as markdown does support image display, while it didn't have native admonition support.

mhils:
Hey there!

Thanks for the extensive write-up. I think images in the docstrings is a reasonable feature request. Some thoughts on this:

[...]

PR #785, opened 2025-03-14
Process HTML `<img>` tags in addition to markdown image markup when embedding local images

earshinov:
Extends #282 and #460.

Demo (copied from extended `testdata/demo_long.py`):

```
<img src="../../docs/logo.png" alt="pdoc logo" width="150">
```

I am interested in this because there is no "official" way to specify image width and height in Markdown markup. `markdown2` doesn't seem to recognize the `pdoc logo` syntax supported by some other Markdown implementations.


test/testdata/demo_long.py
```diff
@@ -39,14 +39,14 @@
 
 FOO_CONSTANT: int = 42
 """
-A happy constant. ✨  
+A happy constant. ✨
```
earshinov, line 42: Sorry, it seems that my VS Code trimmed some trailing whitespaces. Please let me know if you want them back.

earshinov:
Also, not sure why `autofix.ci` introduced so many changes on top of my commit

mhils (commented):
Thanks for the PR! This looks good. Please 1) rebase onto main and 2) add a CHANGELOG entry.

The autofix.ci changes were because you updated uv.lock, which updated our dependencies, which updated the HTML generated by pdoc for snapshot tests. I've done the same in https://github.com/mitmproxy/pdoc/pull/788, so this PR should be clean now.

earshinov:
> Thanks for the PR! This looks good. Please 1) rebase onto main and 2) add a CHANGELOG entry.

Done

earshinov:
> Can you please keep the whitespace changes in CHANGELOG.md at bay? I don't mind the occasional fly-by change, but (some) whitespace in Markdown has semantic meaning.

Sorry, fixed now.

By the way, have you considered using explicit `<br>`'s instead of trailing whitespace to force line breaks?  This way people won't break the layout inadvertently. I know I'm not the first one 🙂

mhils (approved):

mhils:
> By the way, have you considered using explicit `<br>`'s instead of trailing whitespace to force line breaks? This way people won't break the layout inadvertently. I know I'm not the first one 🙂

Subjective preference for Markdown over HTML. I don't feel super strongly, but it's a Markdown document so using Markdown syntax is fine IMO. :)

Thanks again for the PR!

PR #789, opened 2025-03-18
fix numpydoc processing

mhils:
This PR fixes the Numpydoc processing error described in #758.

image

@tobiscode: Does this look good to you?


mhils:
Thanks for the quick feedback loop. My hot take here is that there shouldn't be an expectation of invalid syntax to render nicely. I'll merge this to fix the crash, if you want to shave that yak any further you are welcome to send a PR. 😉

tobiscode:
That's a fair point that there shouldn't be a nice-looking expectation - plus, it'll potentially motivate the user to fix their docstrings ;) Thanks for the quick fix. I think this closes https://github.com/mitmproxy/pdoc/issues/758.

PR #804, opened 2025-05-02
Mark snapshot test artifacts as generated

jinnovation:


mhils (changes requested):
I'd be happy to take both linguist-vendored or linguist-documentation, but the diffs shouldn't be hidden by default.

Issue #825, opened 2025-08-29
Option to add hyperlink third party libraries


Wolf359Stella:
#### Problem Description
Pdoc parses libraries used in the modules documented as modules themselves. These modules, which many times have external documentation in websites could be linked in the same way that the option -e allows liking external code repositories.

#### Proposal
Add a new command line argument that allows the user to pass libraries and hyperlinks using the same interface as -e. The switch could be -l, or link-library. This enhancement could be slowly upgraded, starting with the user needing to manually pass any import string that they want to link, for instance:

  ` -l numpy.ndarray=https://numpy.org/doc/2.3/reference/generated/numpy.ndarray.html `

And later the code could become smart enough to have mapped common used library and try to infer hyperlink using web crawlers.

#### Alternatives
A simpler option is use the output of 
`pip show`
to get a hyperlink for the module

#### Additional context
This helps when a users uses a type hint for functions an methods. Even when common libraries are used, it should not be expect that the type is known by the user, and why not make their life easier by adding a hyperlink to send them directly to the correct web page?

Wolf359Stella:
Based on my investigation, it would be necessary change Function.signature in pdoc/doc.py and format_signature in pdoc/render_helpers.py. I am now investigating how to do this.

Wolf359Stella:
Got it wrong, it was in linkify and link from pdoc/render_helpers.py. I will create a pull request with the enhancement suggest in the way I was able to get it working.

PR #826, opened 2025-08-29
Add hyperlink to both standard libraries and third-party libraries

Wolf359Stella:
Refers to issue #825 .
This pull request implements the following:
* Adds a new command line argument, -l / --link-library, which uses the same format as -e / --edit-url.
* Pass link-library to jinja env as link_library
* Adds the function get_stdlib_doc_link to pdoc/render_helpers.py, which will be used to identify if a string is a reference to a standard python library. If so it returns a hyperlink to it, otherwise it returns None
* Use get_stdlib_doc_link along with link_library dictionary as additional test for type annotation in both linkify and link functions from pdoc/render_helpers.py.

The result is that the documentation now adds hyperlink to both standard libraries and third-party libraries.
I run the pytests and some are failing. I was able to group them into 3 groups:
* Failing because standard libraries now have hyperlinks
* enum.txt, which required to update a substring that looked like it was caused just by python version differences.
* demopackage_dir.html, which I based on search pre-processing and I can´t be sure it is ok.
I fixed all, as asked in the guideline

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ pdoc's main feature is a focus on simplicity: pdoc aims to do one thing and do i
 * Understands numpydoc and Google-style docstrings.
 * Standalone HTML output without additional dependencies.
   
-Under the hood...
+Under the hood... 
 
 * `pdoc` will automatically link identifiers in your docstrings to their corresponding documentation.
 * `pdoc` respects your `__all__` variable when present.


Wolf359Stella:
Hi, I have tried several times but was not able to get all testcases passing. The one failing is on Windows, which I dont use. Could someone help me with this?

PR #832, opened 2025-09-16
Fix linkify of unicode identifiers

### LINKED ISSUE #831: Links to identifiers with unicode characters in them are not created properly
#### Problem Description

When referencing an identifier in backticks that contains a unicode character in the documentation, the resulting link only includes the first parts of the identifier which do not contain a unicode character. For example:

```md
See also: `api_contract.output.rückkauf`
```

Will be rendered as

```html
See also: <code><a href="api_contract/output.html">api_contract.output</a>.rückkauf</code>
```

instead of the expected

```html
See also: <code><a href="api_contract/output/rückkauf.html">api_contract.output.rückkauf</a></code>
```

1. Include the above snippet in your documentation
2. Observe the partial link

#### System Information

```
pdoc: 15.0.1
Python: 3.12.8
Platform: Windows-2019Server-10.0.17763-SP0
```

# Additional context

I already found that this is caused by the Regex in `linkify` using `[a-zA-Z0-9_]` instead of `\w`, which includes Unicode characters as well.

I'll be submitting a PR for this soon.

iFreilicht:
Fixes #831

I'm not sure if there might be a better place for the testdata I added here, but at least it shows what I'm actually fixing.

The desired change in the output can be seen here: https://github.com/mitmproxy/pdoc/pull/832/commits/ec1f716629c8d35381becb55ae080ea6d32d4138#diff-604417b9da1d286441ca229ba9380785833caa22a8ae36263a717c84d9f061c8R1700

### DIFF
--- a/test/testdata/demo_long.html
+++ b/test/testdata/demo_long.html
@@ -182,18 +182,6 @@
             <li>
                     <a class="function" href="#admonitions">admonitions</a>
             </li>
-            <li>
-                    <a class="class" href="#MötörCrüe">MötörCrüe</a>
-                            <ul class="memberlist">
-                        <li>
-                                <a class="function" href="#MötörCrüe.créer_naïveté">créer_naïveté</a>
-                        </li>
-                </ul>
-
-            </li>
-            <li>
-                    <a class="function" href="#link_to_unicode">link_to_unicode</a>
-            </li>
     </ul>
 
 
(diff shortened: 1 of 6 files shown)


mhils (commented):
Thanks! Change looks great, but let's not pollute `demo_long`. Let's add this to `misc` please. :)

PR #841, opened 2025-10-16
fix: remove hard-coded pdoc mention from title in template

### LINKED ISSUE #839: Remove hard-coded `pdoc __version__` from title of module list page
I was reviewing my custom jinja2 templates the other day and found that since #288 I could remove almost all of my customizations. Inline with other configuration options that have been added since then also being exposed via CLI arguments (e.g. #350), I would like to make a case for exposing a `--title` CLI argument to enable easier configuration of this line:
https://github.com/mitmproxy/pdoc/blob/a12c236e35b056cce2603abb40a80c7666823ae1/pdoc/templates/default/index.html.jinja2#L17

Otherwise, any project which does not use a custom `index.html.jinja2` template will claim to contain the module list of pdoc with the `__version__` used to generate the docs, which is obviously not correct. I would suggest to outright remove this part of the title entirely and default to simply using `Module List`. Whether the rest of the string is desirable enough to add a new CLI option for configuring this I think can be up for debate.

mrossinek:
Closes #839 

---

There are two more occurrences of `pdoc {{ __version__ }}` in the jinja2 templates:
1. This one should definitely be kept: https://github.com/mitmproxy/pdoc/blob/a12c236e35b056cce2603abb40a80c7666823ae1/pdoc/templates/default/error.html.jinja2#L36
2. For this one I'm unsure whether it should be kept: https://github.com/mitmproxy/pdoc/blob/a12c236e35b056cce2603abb40a80c7666823ae1/pdoc/templates/default/frame.html.jinja2#L6


mhils (approved):
Thanks! This looks perfect. The other two uses are fine, they are fairly unambiguous.

PR #858, opened 2025-11-11
Support patching docs from imported module stub files

### LINKED ISSUE #791: Stub file information doesn't seem to be included from `_lib.pyi`
#### Package setup
We have some internal namespace package written in Rust & PyO3.

It's organized like this:
```
pyproject.toml
uv.lock
tests/
  [...]
src/
  [internal namespace]/
    [pkg name]/
      __init__.py
      py.typed
      _lib.so
      _lib.pyi
```

`_lib.so` is the shared library built from our Rust code. In our `__init__.py`, we reference this:

```
$ cat __init__.py
from ._lib import [ClassA, ClassB, ...]

__all__ = [
    "[ClassA]",
    "[ClassB]",
    ...
]
```

#### Problem description

When I now create documentation by calling `pdoc [namespace].[pkgname]`, I'm given this function documentation for example:

Image

(ignore the slightly broken formatting of the docstring, this is because we haven't standardized docstring formats yet 😅 )

As you can see, it contains function signature defaults, all parameters and the docstring. But it entirely misses typing information that is in `_lib.pyi`.

When I create documentation with `pdoc [namespace].[pkgname]._lib`, I do see information from the stub file though!

Image

But now, the documentation is limited purely to the `_lib` submodule. Is there a way to create documentation of the entire package, but also correctly discovering stub files automatically?

#### System Information
```
$ pdoc --version
pdoc: 15.0.1
Python: 3.11.9
Platform: Linux-6.13.7-1-MANJARO-x86_64-with-glibc2.41
```

inventshah:
Fixes #791.

Patches docs for variables imported from other modules with external type stubs if the current module has no type stubs. Helps create docs for native extension modules (e.g. from `nanobind` or `PyO3`).

### DIFF
--- a/pdoc/_pydantic.py
+++ b/pdoc/_pydantic.py
@@ -31,7 +31,11 @@ def is_pydantic_model(obj: ClassOrModule) -> TypeGuard[pydantic.BaseModel]:
         # => if we cannot import pydantic, the passed object cannot be a subclass of BaseModel.
         return False
 
-    return isinstance(obj, type) and issubclass(obj, pydantic.BaseModel)
+    return (
+        isinstance(obj, type)
+        and issubclass(obj, pydantic.BaseModel)
+        and obj is not pydantic.BaseModel  # BaseModel doesn't have __pydantic_fields__
+    )
 
 
 def default_value(parent: ClassOrModule, name: str, obj: Any) -> Any:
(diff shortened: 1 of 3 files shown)


inventshah:
Investigating test failures. The 3.14.0 ones looks potentially unrelated, but the others are legit.

PR #872, opened 2026-03-10
Fix external links with anchors being rewritten as relative doc links

mbeijen:
Hi, I tried to setup documentation for a project where I had anchor tags in links in my `README.md` such as http://example.com/foobar#baz  however the linkification module thought these were code references and it would not render properly. Here my naive bugfix, I hope it helps


mhils:
Thank you for the PR! Great suggestion. Can we turn this into a context manager?

```python
with shield_fragments(code, r'<a\s+href="https?://[^"]*"[^>]*>.*?</a>', re.DOTALL) as code:
    ....
```

mbeijen:
Sure, i now implemented it as a context manager, I hope you like it!!

PR #876, opened 2026-03-31
fix: start-after/end-before order in _rst_include_trim

JohanKarlbergg:
Hi,

It looks like `start-after` and `end-before` may be applied in the wrong order. From how I read the docutils spec,  `end-before` is defined as matching content before the first occurrence of the specified text, but still after any `start-after` text. That suggests the current behavior might be reversed, though I may be misunderstanding it.

My use case is to include the text between two markers, for example `"""` and `"""`.

https://docutils.sourceforge.io/docs/ref/rst/directives.html#include-options


mhils (approved):
Thank you! 😃  Do you want to add a CHANGELOG entry?

JohanKarlbergg:
> Thank you! 😃 Do you want to add a CHANGELOG entry?

Sure :)

mhils (approved):
Thanks!

PR #880, opened 2026-04-10
Set ruff target-version to lowest supported python version

clbarnes:
~~The previous `.python-version` = `3.14.0` meant that development environments would automatically be set up using this later python version rather than the oldest supported. This changes it to `3.10.11`, the latest bugfix release of python 3.10, discouraging developers from using unsupported python features.~~

This allows local linting to check for use of unsupported python features.


clbarnes:
We could extend the tox config to run tests or at least basic lints against every supported version.

mhils:
I think 3.14 is the correct choice. We do need to test features that not present in the oldest supported version during development. We do have CI to catch unintended usage of newer features in the main codebase.

clbarnes:
Yes, mypy lacks a per-file target version like ruff has. This was raised here https://github.com/python/mypy/issues/19892 , which was closed in favour of https://github.com/python/mypy/issues/19923 , which was then rejected as impossible (where the original request probably isn't 🙃).

I agree that 3.14 is required for the moment because otherwise mypy will trip up on those misc_py* files. However, I believe that some kind of config to make sure developers don't start using future features and then have to unravel their work based on CI feedback would be welcome. Setting the ruff target-version might do it, I'll do that.

mhils (approved):
Thanks!

